PR #15, opened 2026-02-25
[DCP - Terraform] Initializes the APIs, Service Account, Cloud Run Service and Spanner Instance and DB.

gmechali:
This group of terraform script controls the setup of a new Data Commons Platform deployment within a GCP Project that has nothing setup.

- Enables the Spanner, Cloud Run and IAM APIs
- Creates a Service Account, with databaseUser permission 
- Creates a Cloud Run Service, deploying the :latest datacommons-platform image and grants it AllUsers Invoker permissions
- Creates a Spanner Instance and DB

Note that after running this in datcom-website-dev, we succeeded with each of the following resources:
- DCP Service Account - [🔗](https://pantheon.corp.google.com/iam-admin/serviceaccounts/details/103873044663480688789?e=13803378&inv=1&invt=Ab4PUA&mods=-monitoring_api_staging&project=datcom-website-dev)
- DCP Cloud Run Service - [🔗 ](https://pantheon.corp.google.com/run/detail/us-central1/gmechali-datacommons-platform-test1/observability/metrics?e=13803378&inv=1&invt=Ab4PUA&mods=-monitoring_api_staging&project=datcom-website-dev)
- Spanner Instance [🔗](https://pantheon.corp.google.com/spanner/instances/gmechali-test1-instance/details/databases?e=13803378&inv=1&invt=Ab4PUA&mods=-monitoring_api_staging&project=datcom-website-dev)  and DB [🔗](https://pantheon.corp.google.com/spanner/instances/gmechali-test1-instance/databases/gmechali-test1-db/details/tables?e=13803378&inv=1&invt=Ab4PUA&mods=-monitoring_api_staging&project=datcom-website-dev)

And after running the commands in the DCP setup: https://github.com/datacommonsorg/datacommons?tab=readme-ov-file#2-define-your-schema but changing the URL to the new Cloud Run Service, you can inspect the logs to find the requests, and inspect the DB to find the schema and nodes successfully saved!

Note - I have added more optional variables to control the Spanner DB + cloud run service setup but have not yet tested them. Thoughts on including those? 

Lastly, it adds a setup script to create the GCS bucket for remote state management.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ _override.tf.json
 terraform.rc
 tfplan
 infra/dcp/backend.tf
+.terraform.lock.hcl
 
 # Configuration
 infra/dcp/.env
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,27 @@ node_modules
 .ruff_cache
 .coverage
 htmlcov/
+
+# Terraform
+.terraform/
+*.tfstate
+*.tfstate.*
+crash.log
+crash.log.*
+*.tfvars
+*.tfvars.json
+override.tf
+override.tf.json
+_override.tf
+_override.tf.json
+.terraformrc
+terraform.rc
+tfplan
+infra/dcp/backend.tf
+
+# Configuration
+infra/dcp/.env
+
+*.pyc
+.venv/
+uv.lock
--- a/.gitignore
+++ b/.gitignore
@@ -34,4 +34,3 @@ infra/dcp/.env
 
 *.pyc
 .venv/
-uv.lock
--- a/infra/dcp/README.md
+++ b/infra/dcp/README.md
@@ -1,0 +1,64 @@
+# Data Commons Platform (DCP) Infrastructure
+
+This directory contains the Terraform configuration to deploy the Data Commons Platform on Google Cloud Platform (GCP).
+
+## Prerequisites
+*   **GCP Project**: A GCP project with billing enabled.
+*   **Terraform**: Terraform installed locally (>= 1.0.0).
+*   **gcloud CLI**: GCP CLI installed and authenticated.
+
+## Setup
+
+1.  **Configure Local Variables**:
+    Copy the example variable file and fill in your project details.
+    ```bash
+    cp terraform.tfvars.example terraform.tfvars
+    ```
+    Edit `terraform.tfvars` with your `project_id` and other preferred settings.
+
+2.  **Run Setup Script**:
+    The `setup.sh` script creates a GCS bucket for your Terraform state and initializes the backend.
+    ```bash
+    ./setup.sh
+    ```
+
+## Deployment
+
+1.  **Initialize**:
+    Initialize Terraform (if not already done by setup.sh).
+    ```bash
+    terraform init
+    ```
+
+2.  **Plan**:
+    Review the changes Terraform will make.
+    ```bash
+    terraform plan
+    ```
+
+3.  **Apply**:
+    Provision the infrastructure.
+    ```bash
+    terraform apply
+    ```
+
+4.  **Teardown**:
+    Destroy all resources.
+    ```bash
+    terraform destroy
+    ```
+
+## Architecture
+
+This setup uses an **Orchestrator Pattern**:
+- `infra/dcp/main.tf`: The root entrypoint that calls modules.
+- `infra/dcp/modules/dcp/`: The new Data Commons Platform stack (Cloud Run + Spanner).
+- `infra/dcp/modules/cdc/`: The legacy Custom Data Commons stack (Cloud Run + MySQL + Redis).
+
+Each module is independent and can be toggled via the root variables in `terraform.tfvars`.
+
+## Troubleshooting
+*   **Deletion Errors**: If you get a "cannot destroy... deletion_protection" error, ensure `deletion_protection = false` in your `terraform.tfvars`, run `terraform apply`, and then try `terraform destroy` again. Alternatively, use the helper command:
+    ```bash
+    make force-destroy
+    ```
--- a/infra/dcp/main.tf
+++ b/infra/dcp/main.tf
@@ -1,0 +1,111 @@
+terraform {
+  required_providers {
+    google = {
+      source  = "hashicorp/google"
+      version = ">= 5.0"
+    }
+    null = {
+      source  = "hashicorp/null"
+      version = ">= 3.0"
+    }
+  }
+}
+
+provider "google" {
+  project = var.project_id
+  region  = var.region
+}
+
+# Enable required APIs for both stacks
+resource "google_project_service" "apis" {
+  for_each = toset([
+    "run.googleapis.com",
+    "spanner.googleapis.com",
+    "iam.googleapis.com",
+    "sqladmin.googleapis.com",
+    "redis.googleapis.com",
+    "secretmanager.googleapis.com",
+    "vpcaccess.googleapis.com",
+    "artifactregistry.googleapis.com",
+    "compute.googleapis.com"
+  ])
+
+  service            = each.key
+  disable_on_destroy = false
+}
+
+# --- Data Commons Platform (DCP) Stack ---
+module "dcp" {
+  source = "./modules/dcp"
+  count  = var.enable_dcp ? 1 : 0
+
+  project_id               = var.project_id
+  region                   = var.region
+  image_url                = var.dcp_image_url
+  service_name             = var.dcp_service_name
+  service_account_name     = var.dcp_service_account_name
+  create_spanner           = var.dcp_create_spanner
+  spanner_instance_id      = var.dcp_spanner_instance_id
+  spanner_database_id      = var.dcp_spanner_database_id
+  spanner_processing_units = var.dcp_spanner_processing_units
+  cpu                      = var.dcp_cpu
+  memory                   = var.dcp_memory
+  min_instances            = var.dcp_min_instances
+  max_instances            = var.dcp_max_instances
+  concurrency              = var.dcp_concurrency
+  timeout_seconds          = var.dcp_timeout_seconds
+  deletion_protection      = var.deletion_protection
+
+  depends_on = [google_project_service.apis]
+}
+
+# --- Custom Data Commons (CDC) Legacy Stack ---
+module "cdc" {
+  source = "./modules/cdc"
+  count  = var.enable_cdc ? 1 : 0
+
+  project_id                    = var.project_id
+  namespace                     = var.cdc_namespace
+  dc_api_key                    = var.cdc_dc_api_key
+  maps_api_key                  = var.cdc_maps_api_key
+  disable_google_maps           = var.cdc_disable_google_maps
+  region                        = var.region
+  google_analytics_tag_id       = var.cdc_google_analytics_tag_id
+  gcs_data_bucket_name          = var.cdc_gcs_data_bucket_name
+  gcs_data_bucket_input_folder  = var.cdc_gcs_data_bucket_input_folder
+  gcs_data_bucket_output_folder = var.cdc_gcs_data_bucket_output_folder
+  gcs_data_bucket_location      = var.cdc_gcs_data_bucket_location
+  mysql_instance_name           = var.cdc_mysql_instance_name
+  mysql_database_name           = var.cdc_mysql_database_name
+  mysql_database_version        = var.cdc_mysql_database_version
+  mysql_cpu_count               = var.cdc_mysql_cpu_count
+  mysql_memory_size_mb          = var.cdc_mysql_memory_size_mb
+  mysql_storage_size_gb         = var.cdc_mysql_storage_size_gb
+  mysql_user                    = var.cdc_mysql_user
+  mysql_deletion_protection     = var.deletion_protection
+  dc_web_service_image          = var.cdc_web_service_image
+  dc_web_service_min_instance_count = var.cdc_web_service_min_instance_count
+  dc_web_service_max_instance_count = var.cdc_web_service_max_instance_count
+  dc_web_service_cpu            = var.cdc_web_service_cpu
+  dc_web_service_memory         = var.cdc_web_service_memory
+  make_dc_web_service_public    = var.cdc_make_dc_web_service_public
+  dc_data_job_image             = var.cdc_data_job_image
+  dc_data_job_cpu               = var.cdc_data_job_cpu
+  dc_data_job_memory            = var.cdc_data_job_memory
+  dc_data_job_timeout           = var.cdc_data_job_timeout
+  dc_search_scope               = var.cdc_search_scope
+  enable_mcp                   = var.cdc_enable_mcp
+  vpc_network_name              = var.cdc_vpc_network_name
+  vpc_network_subnet_name       = var.cdc_vpc_network_subnet_name
+  enable_redis                  = var.cdc_enable_redis
+  redis_instance_name           = var.cdc_redis_instance_name
+  redis_memory_size_gb          = var.cdc_redis_memory_size_gb
+  redis_tier                    = var.cdc_redis_tier
+  redis_location_id             = var.cdc_redis_location_id
+  redis_alternative_location_id = var.cdc_redis_alternative_location_id
+  redis_replica_count           = var.cdc_redis_replica_count
+  vpc_connector_cidr            = var.cdc_vpc_connector_cidr
+  deletion_protection           = var.deletion_protection
+
+  depends_on = [google_project_service.apis]
+}
--- a/infra/dcp/main.tf
+++ b/infra/dcp/main.tf
@@ -12,13 +12,16 @@ terraform {
 }
 
 provider "google" {
-  project = var.project_id
-  region  = var.region
+  project               = var.project_id
+  region                = var.region
+  user_project_override = var.user_project_override
+  billing_project       = var.billing_project_id != null ? var.billing_project_id : var.project_id
 }
 
 # Enable required APIs for both stacks
 resource "google_project_service" "apis" {
   for_each = toset(concat([
+    "apikeys.googleapis.com",
     "run.googleapis.com",
     "iam.googleapis.com",
     "sqladmin.googleapis.com",
@@ -31,6 +34,16 @@ resource "google_project_service" "apis" {
 
   service            = each.key
   disable_on_destroy = false
+}
+
+# --- Network Data Sources ---
+data "google_compute_network" "default" {
+  name = var.cdc_vpc_network_name
+}
+
+data "google_compute_subnetwork" "default" {
+  name   = var.cdc_vpc_network_subnet_name
+  region = var.region
 }
 
 # --- Data Commons Platform (DCP) Stack ---
@@ -106,6 +119,7 @@ module "cdc" {
   redis_alternative_location_id = var.cdc_redis_alternative_location_id
   redis_replica_count           = var.cdc_redis_replica_count
   vpc_connector_cidr            = var.cdc_vpc_connector_cidr
+  vpc_network_id                = data.google_compute_network.default.id
   use_spanner                   = var.enable_dcp
   spanner_instance_id           = var.enable_dcp ? module.dcp[0].spanner_instance_id : ""
   spanner_database_id           = var.enable_dcp ? module.dcp[0].spanner_database_id : ""
--- a/infra/dcp/main.tf
+++ b/infra/dcp/main.tf
@@ -18,9 +18,8 @@ provider "google" {
 
 # Enable required APIs for both stacks
 resource "google_project_service" "apis" {
-  for_each = toset([
+  for_each = toset(concat([
     "run.googleapis.com",
-    "spanner.googleapis.com",
     "iam.googleapis.com",
     "sqladmin.googleapis.com",
     "redis.googleapis.com",
@@ -28,7 +27,7 @@ resource "google_project_service" "apis" {
     "vpcaccess.googleapis.com",
     "artifactregistry.googleapis.com",
     "compute.googleapis.com"
-  ])
+  ], var.enable_dcp ? ["spanner.googleapis.com"] : []))
 
   service            = each.key
   disable_on_destroy = false
@@ -118,7 +117,7 @@ module "cdc" {
 # Ensure Spanner instance ID is provided when not creating a new one
 check "spanner_instance_id_provided" {
   assert {
-    condition     = var.dcp_create_spanner_instance || var.dcp_spanner_instance_id != ""
-    error_message = "dcp_spanner_instance_id must be provided when dcp_create_spanner_instance is false."
+    condition     = !var.enable_dcp || var.dcp_create_spanner_instance || var.dcp_spanner_instance_id != ""
+    error_message = "dcp_spanner_instance_id must be provided when reusing an existing instance (dcp_create_spanner_instance = false)."
   }
 }
--- a/infra/dcp/main.tf
+++ b/infra/dcp/main.tf
@@ -40,20 +40,22 @@ module "dcp" {
   count  = var.enable_dcp ? 1 : 0
 
   project_id               = var.project_id
+  namespace                = var.namespace
   region                   = var.region
   image_url                = var.dcp_image_url
   service_name             = var.dcp_service_name
   service_account_name     = var.dcp_service_account_name
-  create_spanner           = var.dcp_create_spanner
+  create_spanner_instance  = var.dcp_create_spanner_instance
+  create_spanner_db        = var.dcp_create_spanner_db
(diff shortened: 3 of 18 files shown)


infra/dcp/variables.tf
```diff
@@ -0,0 +1,338 @@
...
+  description = "Enable deletion protection for resources (set to true for production)"
+  type        = bool
+  default     = false
+}
+
+# --- Stack Toggles ---
+variable "enable_dcp" {
+  description = "Enable the new Data Commons Platform stack"
+  type        = bool
+  default     = true
```
dwnoble: I think this should default to `false` until we're ready to launch. Enabling dcp also implies enabling spanner and thus dropping support for the v0 & v1 mixer apis. We've still got some work to do to migrate the website off of the v0 and v1 apis.

infra/dcp/variables.tf
```diff
@@ -0,0 +1,338 @@
...
+}
+
+variable "enable_cdc" {
+  description = "Enable the legacy Custom Data Commons stack"
+  type        = bool
+  default     = false
```
dwnoble: For backwards compatibility i think we should defaul this to `true`

infra/dcp/Makefile
```diff
@@ -0,0 +1,34 @@
+# Makefile for Data Commons Infrastructure
+
+.PHONY: init plan apply destroy fmt validate
+
+# Initialize terraform
```
dwnoble: what is the value added by this makefile? it looks like most of the commands are pass through to single terraform commands

infra/dcp/terraform.tfvars.example
```diff
@@ -0,0 +1,20 @@
+# --- Shared Global Variables ---
+project_id          = "your-project-id"
+region              = "us-central1"
+deletion_protection = false
```
dwnoble: I'd remove variables where we are already setting default values

infra/dcp/terraform.tfvars.example
```diff
@@ -0,0 +1,20 @@
...
+# --- DCP Stack Variables ---
+dcp_image_url           = "gcr.io/your-project/datacommons-platform:latest"
+dcp_service_name        = "datacommons-platform"
+dcp_service_account_name = "dcp-runner-sa"
+dcp_create_spanner       = true
+
+# --- CDC Stack Variables (Legacy) ---
+# cdc_namespace         = "cdc"
+# cdc_dc_api_key        = "your-dc-api-key"
+# cdc_maps_api_key      = "your-maps-api-key"
```
dwnoble: the maps key gets auto-generated if it's undefined, so i would leave it out

infra/dcp/modules/dcp/cloudrun.tf
```diff
@@ -0,0 +1,51 @@
+resource "google_cloud_run_v2_service" "dcp_service" {
+  name                = var.service_name
```
dwnoble: can you add the `${var.namespace}-` prefix to the created dcp assets? we can reuse the CDC namespace variable. this lets us deploy multiple CDC+DCP instances in the same GCP account without having name collisions

infra/dcp/variables.tf
```diff
@@ -0,0 +1,338 @@
...
+variable "dcp_service_name" {
+  description = "Cloud Run service name for DCP"
+  type        = string
+  default     = "datacommons-platform"
+}
+
+variable "dcp_service_account_name" {
+  description = "Service account for DCP"
+  type        = string
+  default     = "dcp-runner-sa-testing123"
```
dwnoble: Nit: rename to dcp-runner-sa

infra/dcp/variables.tf
```diff
@@ -0,0 +1,338 @@
...
+}
+
+variable "dcp_create_spanner" {
+  description = "Create Spanner for DCP"
+  type        = bool
+  default     = false
```
dwnoble, line 66: I would only use this variable to control if a spanner _instance_ gets created, but we can still create the spanner database with terraform. This is mostly for cost savings - creating new instances is expensive, but creating new databases is cheap. telling terraform not to create the instance means several dcp instances can reuse the same instance.

infra/dcp/variables.tf
```diff
@@ -0,0 +1,338 @@
...
+}
+
+variable "dcp_spanner_instance_id" {
+  description = "Spanner instance for DCP"
+  type        = string
+  default     = "dcp-spanner-inst-test123"
```
dwnoble: Nit: remove test123

infra/dcp/variables.tf
```diff
@@ -0,0 +1,338 @@
...
+  default     = "dcp-spanner-db-test123"
+}
+
+variable "dcp_spanner_processing_units" {
+  description = "Spanner units for DCP"
+  type        = number
+  default     = 100
+}
+
+variable "dcp_cpu" {
```
dwnoble: Nit: Make it clear that this variable is for the dcp service container cpu. Granted that's the only service container now, but i could see us adding more containers in the future.

infra/dcp/variables.tf
```diff
@@ -0,0 +1,338 @@
...
+}
+
+variable "dcp_timeout_seconds" {
+  description = "DCP timeout"
+  type        = number
+  default     = 300
+}
+
+# --- CDC Stack Variables (Legacy) ---
+variable "cdc_namespace" {
```
dwnoble: Can you make this namespace apply to both cdc & dcp?

infra/dcp/.env.example
```diff
@@ -0,0 +1,28 @@
+# GCP Project ID
```
dwnoble: Are the .env files still necessary now that you've added terraform.tfvars files?

infra/dcp/.terraform.lock.hcl
```diff
@@ -0,0 +1,61 @@
+# This file is maintained automatically by "terraform init".
```
dwnoble: Nit: rm this file and add it to .gitignore

infra/dcp/terraform.tfvars.example
```diff
@@ -0,0 +1,20 @@
...
+enable_cdc          = false
+
+# --- DCP Stack Variables ---
+dcp_image_url           = "gcr.io/your-project/datacommons-platform:latest"
+dcp_service_name        = "datacommons-platform"
+dcp_service_account_name = "dcp-runner-sa"
+dcp_create_spanner       = true
+
+# --- CDC Stack Variables (Legacy) ---
+# cdc_namespace         = "cdc"
```
dwnoble: For backwards compatibility, we should consider keeping the cdc variables named without the cdc_* prefix. this way existing terraform users can start using dcp by just setting `enable_dcp = true`

infra/dcp/modules/cdc/locals.tf
```diff
@@ -0,0 +1,90 @@
...
+
+  # Optionally-configured Redis instance
+  redis_instance = var.enable_redis ? google_redis_instance.redis_instance[0] : null
+
+
+  # Shared environment variables used by the Data Commons web service and the Data
+  # Commons data loading job
+  cloud_run_shared_env_variables = [
+    {
+      name  = "USE_CLOUDSQL"
```
dwnoble, line 28: We'll also have to update CDC to drop the cloudsql configuration and instead use a spanner configuration when dcp is enabled

infra/dcp/terraform.tfvars.example
gmechali: Do you know where is it auto generated?

infra/dcp/Makefile
gmechali: removed - I think it just gave the framework for if we needed anything more complex. 
I had the force destroy command butprobably overkill for now. Removed.

infra/dcp/variables.tf
gmechali: done.

infra/dcp/variables.tf
gmechali: Done.

infra/dcp/variables.tf
gmechali: done.

infra/dcp/variables.tf
gmechali: done.

infra/dcp/.env.example
gmechali: Nope - removed

infra/dcp/variables.tf
gmechali: so is this namespace meant to be like "un_ilo"? or "one" or "techsoup"

And then everything will get prefixed as "techsoup_dcp_service"?

infra/dcp/.terraform.lock.hcl
gmechali: done.

infra/dcp/terraform.tfvars.example
gmechali: done

infra/dcp/modules/dcp/cloudrun.tf
gmechali: Done.

infra/dcp/terraform.tfvars.example
gmechali: I want to think a bit more about backwards compatibility. I want to see if we can somehow onboard onto this terraform the ones setup already. 

Let me take a stab at this tomorrow outside of the scope of this PR :)

infra/dcp/modules/cdc/locals.tf
gmechali: Done

infra/dcp/terraform.tfvars.example
gmechali: I think this is how we handle backwards compatibility: https://developer.hashicorp.com/terraform/language/block/import
Still more investigation to do, but leaving for a later PR

infra/dcp/terraform.tfvars.example
dwnoble: Here: https://github.com/datacommonsorg/website/blob/master/deploy/terraform-custom-datacommons/modules/main.tf#L139

Using this tf module: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/apikeys_key

infra/dcp/variables.tf
dwnoble: yeah- it may aalso be "un-ilo-staging" or "un-ilo-production" to let customers host multiple environments in the same account

infra/dcp/setup.sh
```diff
@@ -0,0 +1,72 @@
...
+echo ""
+
+# Check for terraform.tfvars file
+if [ ! -f terraform.tfvars ]; then
+    echo "⚠️  Warning: No terraform.tfvars file found. You might want to copy terraform.tfvars.example to terraform.tfvars first."
+fi
+
+# Prompt for bucket name
+read -p "Enter the name of your GCS Terraform State Bucket: " BUCKET_NAME
+
```
dwnoble, line 18: Nit: I ran this script and it wasn't clear to me which GCP project the bucket would be created in. Can we do the project identification step first and print out the project that we're doing setup in before creating the bucket as a sanity check?

infra/dcp/terraform.tfvars.example
```diff
@@ -0,0 +1,15 @@
+# --- Shared Global Variables ---
+project_id          = "your-project-id"
+namespace           = "techsoup"
+
+# --- Stack Toggles ---
+enable_dcp                  = true
+enable_cdc                  = false
```
dwnoble: Nit: set enable_cdc to true

infra/dcp/terraform.tfvars.example
```diff
@@ -0,0 +1,15 @@
...
+
+# Only set to true if you need to create a dedicated Spanner Instance (Expensive!)
+# Otherwise, provide an existing instance ID in dcp_spanner_instance_id
+dcp_create_spanner_instance = false
+
+# --- CDC Stack Variables (Legacy) ---
+# cdc_dc_api_key        = "your-dc-api-key"
```
dwnoble: Nit: mention here (and/or in the readme) to get an api key from apikeys.datacommons.org

infra/dcp/README.md
```diff
@@ -0,0 +1,64 @@
+# Data Commons Platform (DCP) Infrastructure
+
+This directory contains the Terraform configuration to deploy the Data Commons Platform on Google Cloud Platform (GCP).
+
+## Prerequisites
+*   **GCP Project**: A GCP project with billing enabled.
+*   **Terraform**: Terraform installed locally (>= 1.0.0).
+*   **gcloud CLI**: GCP CLI installed and authenticated.
+
+## Setup
```
clincoln8, line 10: do users need to clone the repo? Is there a command/script we can provide such that they might pull the required files with curl instead having to clone? Curious to hear your general thoughts on this / what's feasible.

Like maybe we have a script that does the pulling of files. So user flow might look like:
1. curl / pull "download_script"
2. run "download_script" which downloads the rest of the required files
3. edit variable file
4. run "setup_script"

infra/dcp/modules/cdc/main.tf
```diff
@@ -0,0 +1,342 @@
...
+
+# Custom Data Commons terraform resources
+
+# Reference the default VPC network
+data "google_compute_network" "default" {
+  name = var.vpc_network_name
+}
+
+# Reference the default VPC network subnet
+data "google_compute_subnetwork" "default" {
```
dwnoble: Can you move the google_compute_network and google_compute_subnetwork outside of the CDC module & up to the top-level main.tf?

With the DCP & CDC components modularized, i think it makes sense to leave network management owned by the top-level module.

infra/dcp/modules/cdc/variables.tf
```diff
@@ -0,0 +1,266 @@
...
+}
+
+variable "redis_alternative_location_id" {
+  description = "Redis alternate location id (alternate zone)"
+  type        = string
+  default     = "us-central1-b"
+}
+
+variable "redis_replica_count" {
+  description = "Redis reserved IP range"
```
dwnoble: Nit: fix description

infra/dcp/modules/cdc/variables.tf
```diff
@@ -0,0 +1,266 @@
...
+  type        = number
+  default     = 1
+}
+
+variable "vpc_connector_cidr" {
+  description = "CIDR range for the VPC Access Connector"
+  type        = string
+  default     = "10.8.0.0/28"
```
dwnoble: Can you set this to the same default as the cidr range in infra/dcp/variables.tf?

infra/dcp/terraform.tfvars.example
```diff
@@ -0,0 +1,15 @@
...
+
+# --- Stack Toggles ---
+enable_dcp                  = true
+enable_cdc                  = false
+
+# Only set to true if you need to create a dedicated Spanner Instance (Expensive!)
+# Otherwise, provide an existing instance ID in dcp_spanner_instance_id
+dcp_create_spanner_instance = false
+
+# --- CDC Stack Variables (Legacy) ---
```
dwnoble, line 13: Can you add a commented out `cdc_vpc_connector_cidr` variable here? when deploying to a shared account we'll need to set that variable to avoid conflicts with other deployments

infra/dcp/main.tf
```diff
@@ -0,0 +1,123 @@
...
+    null = {
+      source  = "hashicorp/null"
+      version = ">= 3.0"
+    }
+  }
+}
+
+provider "google" {
+  project = var.project_id
+  region  = var.region
```
dwnoble: I had to add the billing account to the provider to get terraform apply to run.

I tried setting the billing account manually using `gcloud auth application-default set-quota-project datcom-website-prod`, but i kept getting the error:
```
│ Error: Error creating Key: failed to create a diff: failed to retrieve Key resource: googleapi: Error 403: Your application is authenticating by using local Application Default Credentials. The apikeys.googleapis.com API requires a quota project, which is not set by default. To learn how to set your quota project, see https://cloud.google.com/docs/authentication/adc-troubleshooting/user-creds .
```

Adding these two lines to `provider "google" fixed it:
```
 user_project_override = var.user_project_override
 billing_project       = var.billing_project_id != null ? var.billing_project_id : var.project_id
```


I also added the billing_project_id and user_project_override to infra/dcp/variables.tf :

```

variable user_project_override {
  description = "Set to true to specify a quota / billing project with billing_project_id. Default: true."
  type        = bool
  default     = true
}

variable billing_project_id {
  description = "If user_project_override is set to true, will use this billing project id. Default: null (will use var.project_id as the billing project)"
  type        = bool
  default     = null
}
```

infra/dcp/modules/cdc/main.tf
```diff
@@ -0,0 +1,342 @@
...
+  name   = var.vpc_network_subnet_name
+  region = var.region
+}
+
+# Enable required Google Cloud APIs
+resource "google_project_service" "required_apis" {
```
dwnoble: Can you consolidate this with the other `"google_project_service" "apis"` definition?

infra/dcp/modules/cdc/main.tf
```diff
@@ -0,0 +1,342 @@
...
+  for_each = toset([
+    "run.googleapis.com",
+    "sqladmin.googleapis.com",
+    "compute.googleapis.com",
+    "redis.googleapis.com",
+    "secretmanager.googleapis.com",
+    "vpcaccess.googleapis.com",
+    "artifactregistry.googleapis.com",
+    "iam.googleapis.com"
```
dwnoble: I had to add `"apikeys.googleapis.com"` here to get terraform to generate the maps api key

infra/dcp/variables.tf
```diff
@@ -0,0 +1,343 @@
...
+variable "dcp_service_memory" {
+  description = "Memory limit for the DCP service container"
+  type        = string
+  default     = "512Mi"
+}
+
+variable "dcp_service_min_instances" {
+  description = "Minimum number of instances for the DCP service"
+  type        = number
+  default     = 0
```
dwnoble: Set min instances to 1 so the service is always running

infra/dcp/modules/cdc/main.tf
```diff
@@ -0,0 +1,342 @@
...
+resource "google_apikeys_key" "maps_api_key" {
+  count        = var.maps_api_key == null && !var.disable_google_maps ? 1 : 0
+  name         = "${local.name_prefix}maps-key"
+  display_name = "Maps API Key for ${var.namespace != "" ? var.namespace : "Data Commons"}"
+  project      = var.project_id
+
+  restrictions {
+    api_targets {
+      service = "maps-backend.googleapis.com"
+    }
```
dwnoble, line 120: Add 
```
   api_targets {
     service = "places_backend"
   }
```

so place autocomplete on the visualization tools works

infra/dcp/modules/cdc/main.tf
```diff
@@ -0,0 +1,342 @@
...
+      vpc_access {
+        connector = google_vpc_access_connector.connector.id
+        egress    = "PRIVATE_RANGES_ONLY"
+      }
+      max_retries = 0
+      timeout     = var.dc_data_job_timeout
+      service_account = google_service_account.datacommons_service_account.email
+    }
+  }
+  depends_on = [google_project_service.required_apis]
```
dwnoble: I had to change this dependency list to:
```
depends_on = [
    google_secret_manager_secret_version.mysql_password_secret_version,
    google_secret_manager_secret_version.dc_api_key_version,
    google_secret_manager_secret_version.maps_api_key_version
  ]
```

to get terraform to create the resource without errors

infra/dcp/modules/cdc/main.tf
```diff
@@ -0,0 +1,342 @@
...
+}
+
+# Run the db init job on terraform apply to create tables
+resource "null_resource" "run_db_init" {
```
dwnoble, line 199: This db init script only applies to mysql configurations. Add

```
 count            = var.use_spanner ? 0 : 1
```

dwnoble (changes requested):
Thanks Gabe! I had to make a few changes to the tf configuration to get it working on my end, but with these changes my cdc+dcp stacks are up and running

infra/dcp/setup.sh
gmechali: I guess it should probably be optional to connect to remote state management. Made it optional,

If you wanted to connect to the changes I deployed, you should use datacommons-platform-tf-state-1

infra/dcp/terraform.tfvars.example
gmechali: done.

infra/dcp/modules/cdc/main.tf
gmechali: done.

infra/dcp/variables.tf
gmechali: done.

infra/dcp/modules/cdc/main.tf
gmechali: Done

infra/dcp/README.md
gmechali: We had a conversation about where this should live. For now I ll get this checked in here, but it's an ongoing conversation of whether we will move it to a Terraform Data Commons Repo.

This PR is pretty isolated so it will be very simple to refactor. I think we will get the conversationsettled shortly :)

infra/dcp/modules/cdc/variables.tf
gmechali: ah good catch thanks.

infra/dcp/terraform.tfvars.example
gmechali: done.

infra/dcp/modules/cdc/main.tf
gmechali: done.

infra/dcp/modules/cdc/main.tf
gmechali: fixed

infra/dcp/modules/cdc/main.tf
gmechali: hm place autoocomplete on vistools works for me...
https://screenshot.googleplex.com/giJCCSSEUqiq8iV

What is places_backend referencing?

infra/dcp/modules/cdc/main.tf
gmechali: done.

infra/dcp/main.tf
gmechali: done.

.gitignore
```diff
@@ -9,4 +9,16 @@ venv
...
+.terraform/
+*.tfstate
+*.tfstate.backup
+
+# Configuration
+infra/dcp/.env
+
+*.pyc
+.venv/
+uv.lock
```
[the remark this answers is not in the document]
dwnoble: This comment is right-can you remove uv.lock from the .gitignore?

infra/dcp/modules/dcp/iam.tf
```diff
@@ -0,0 +1,22 @@
...
+}
+
+# Make the Cloud Run service public (optional, can be restricted)
+resource "google_cloud_run_service_iam_binding" "public_invoker" {
+  location = google_cloud_run_v2_service.dcp_service.location
+  service  = google_cloud_run_v2_service.dcp_service.name
+  role     = "roles/run.invoker"
+  members = [
+    "allUsers"
+  ]
```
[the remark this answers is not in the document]
dwnoble: Can you add a similar check to only set public access if `make_dc_web_service_public` is true like in `infra/dcp/modules/cdc/main.tf` ?

infra/dcp/modules/cdc/main.tf
dwnoble: https://screenshot.googleplex.com/3TxzHxMnXBzZCY9Hmm i get an error with the places backend disabled: 

you may be using an unrestricted key: https://screenshot.googleplex.com/AHHtecWh6pKZVd2

dwnoble (changes requested):

infra/dcp/modules/cdc/main.tf
gmechali: Ah I see, thanks for catching this! SHould be fixed!

.gitignore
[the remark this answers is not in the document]
gmechali: done.

infra/dcp/modules/dcp/iam.tf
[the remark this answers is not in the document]
gmechali: hey! Im going to deal with this in a follow up. We also cant really test it because of our GCP corp policies.
Took on a task to follow up on this.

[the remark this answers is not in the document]
dwnoble: sgtm!

dwnoble (approved):

PR #16, opened 2026-03-06
Updated data ingestion to use spanner API

dwnoble:
Refactors graph node ingestion to use native Spanner batch mutations for performance, handles Spanner's index key limits with bytes storage pattern, and adds tests for `GraphService`

* Spanner Batch Inserts: Uses native spanner.Batch instead of SQLAlchemy for bulk insertions.
* Byte Limit Handling: Truncates object_value to 4096 bytes to stay within Spanner's 8192-byte index limit. The full payload is preserved in a new object_bytes column.
* Model Updates: Added object_bytes column, extracted table name constants, and removed FetchedValue().
* API Routing: Added trailing slash support (/nodes/) to GET/POST endpoints.
* Bugfixes & Cleanup: Stripped CURIE namespaces before DB insertion. Suppressed noisy telemetry logs and removed old sqlparse hacks. Added google-cloud-spanner to dependencies.
* Testing: Added test suite for `GraphService` with mocked SQLAlchemy and Spanner clients.

### DIFF
--- a/packages/datacommons-api/datacommons_api/api_cli.py
+++ b/packages/datacommons-api/datacommons_api/api_cli.py
@@ -77,8 +77,12 @@ def start(
 
 @api.command()
 @click.option("--gcp-project-id", help="GCP project id.", required=True)
-@click.option("--gcp-spanner-instance-id", help="GCP Spanner instance id.", required=True)
-@click.option("--gcp-spanner-database-name", help="GCP Spanner database name.", required=True)
+@click.option(
+    "--gcp-spanner-instance-id", help="GCP Spanner instance id.", required=True
+)
+@click.option(
+    "--gcp-spanner-database-name", help="GCP Spanner database name.", required=True
+)
 def drop_tables(
     gcp_project_id: str,
     gcp_spanner_instance_id: str,
--- a/packages/datacommons-api/datacommons_api/api_cli.py
+++ b/packages/datacommons-api/datacommons_api/api_cli.py
@@ -91,6 +91,7 @@ def drop_tables(
     yes: bool,
 ):
     """Drop Node and Edge tables from the graph database."""
+    # TODO: Refactor this method to only drop the data from the tables, not the tables themselves.
     if not yes:
         click.confirm(
             "Are you sure you want to drop the Node and Edge tables?", abort=True
--- a/packages/datacommons-api/datacommons_api/api_cli.py
+++ b/packages/datacommons-api/datacommons_api/api_cli.py
@@ -16,9 +16,10 @@ import click
 import uvicorn
 
 from datacommons_api.app import app
-from datacommons_api.core.config import initialize_config
+from datacommons_api.core.config import get_config, initialize_config
 from datacommons_api.core.logging import get_logger, setup_logging
-from datacommons_db.session import initialize_db
+from datacommons_db.session import get_session, initialize_db
+from datacommons_api.services.graph_service import GraphService
 
 setup_logging()
 logger = get_logger(__name__)
@@ -72,3 +73,30 @@ def start(
         port=port,
         reload=reload,
     )
+
+
+@api.command()
+@click.option("--gcp-project-id", help="GCP project id.", required=True)
+@click.option("--gcp-spanner-instance-id", help="GCP Spanner instance id.", required=True)
+@click.option("--gcp-spanner-database-name", help="GCP Spanner database name.", required=True)
+def drop_tables(
+    gcp_project_id: str,
+    gcp_spanner_instance_id: str,
+    gcp_spanner_database_name: str,
+):
+    """Drop Node and Edge tables from the graph database."""
+    logger.info("Dropping Node and Edge tables from the graph database")
+    initialize_config(
+        gcp_project_id=gcp_project_id,
+        gcp_spanner_instance_id=gcp_spanner_instance_id,
+        gcp_spanner_database_name=gcp_spanner_database_name,
+    )
+    config = get_config()
+    db = get_session(
+        config.GCP_PROJECT_ID,
+        config.GCP_SPANNER_INSTANCE_ID,
+        config.GCP_SPANNER_DATABASE_NAME,
+    )
+    graph_service = GraphService(db)
+    graph_service.drop_tables()
+    logger.info("Successfully dropped Node and Edge tables")
--- a/packages/datacommons-api/datacommons_api/api_cli.py
+++ b/packages/datacommons-api/datacommons_api/api_cli.py
@@ -83,9 +83,7 @@ def start(
 @click.option(
     "--gcp-spanner-database-name", help="GCP Spanner database name.", required=True
 )
-@click.option(
-    "--yes", is_flag=True, help="Skip confirmation prompt."
-)
+@click.option("--yes", is_flag=True, help="Skip confirmation prompt.")
 def drop_tables(
     gcp_project_id: str,
     gcp_spanner_instance_id: str,
@@ -94,7 +92,9 @@ def drop_tables(
 ):
     """Drop Node and Edge tables from the graph database."""
     if not yes:
-        click.confirm("Are you sure you want to drop the Node and Edge tables?", abort=True)
+        click.confirm(
+            "Are you sure you want to drop the Node and Edge tables?", abort=True
+        )
 
     logger.info("Dropping Node and Edge tables from the graph database")
     initialize_config(
--- a/packages/datacommons-api/datacommons_api/api_cli.py
+++ b/packages/datacommons-api/datacommons_api/api_cli.py
@@ -83,12 +83,19 @@ def start(
 @click.option(
     "--gcp-spanner-database-name", help="GCP Spanner database name.", required=True
 )
+@click.option(
+    "--yes", is_flag=True, help="Skip confirmation prompt."
+)
 def drop_tables(
     gcp_project_id: str,
     gcp_spanner_instance_id: str,
     gcp_spanner_database_name: str,
+    yes: bool,
 ):
     """Drop Node and Edge tables from the graph database."""
+    if not yes:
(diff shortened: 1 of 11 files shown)


packages/datacommons-api/datacommons_api/services/graph_service.py
```diff
@@ -203,6 +228,140 @@ def node_model_to_graph_node(node: NodeModel) -> GraphNode:
...
+        # Add node and its edges to the current batch
+        node_len = len(node_model.outgoing_edges) + 1
+        if current_batch_len + node_len < batch_size:
+            current_batch.append(node_model)
+            current_batch_len += node_len
+        else:
+            # If the current batch is full, add it to the list of batches
+            node_batches.append(current_batch)
+            current_batch = []
+            current_batch_len = 0
```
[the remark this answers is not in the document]
dwnoble: Fixed

packages/datacommons-api/datacommons_api/services/graph_service.py
```diff
@@ -203,6 +228,149 @@ def node_model_to_graph_node(node: NodeModel) -> GraphNode:
...
+    Helper function to get the value of an edge column, with support for Spanner index key length limits.
+
+    Args:
+      e: The EdgeModel instance
+      col: The column name
+
+    Returns:
+      The value of the edge column, with support for Spanner index key length limits
+    """
+    if col in ("object_value", "object_bytes"):
```
clincoln8: nit: return early for the simple case of getting a non object_value/object_bytes col
```suggestion
    if col not in ("object_value", "object_bytes"):
      return getattr(e, col)
    val = ...
```

packages/datacommons-api/datacommons_api/services/graph_service.py
```diff
@@ -203,6 +228,140 @@ def node_model_to_graph_node(node: NodeModel) -> GraphNode:
...
+                val_truncated = val_bytes[:OBJECT_VALUE_MAX_LENGTH].decode(
+                    "utf-8", errors="ignore"
+                )
+                return val_truncated
+            return val
+        elif col == "object_bytes":
+            if len(val_bytes) > OBJECT_VALUE_MAX_LENGTH:
+                import base64
+
+                return base64.b64encode(val_bytes).decode("utf-8")
```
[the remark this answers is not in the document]
clincoln8: +1, i'm not sure why we'd decode the object_bytes value

packages/datacommons-api/datacommons_api/services/graph_service.py
```diff
@@ -203,6 +228,149 @@ def node_model_to_graph_node(node: NodeModel) -> GraphNode:
...
+    """
+    Helper function to get the value of an edge column, with support for Spanner index key length limits.
+
+    Args:
+      e: The EdgeModel instance
+      col: The column name
+
+    Returns:
+      The value of the edge column, with support for Spanner index key length limits
+    """
```
clincoln8: nit on method name and docstring -- feel free to ignore
```suggestion
def coerce_edge_val_for_db_write(e: EdgeModel, col: str) -> str | None:
    """
    Coerces and truncates edge values to comply with Spanner index limits.

    Args:
      e: The EdgeModel instance containing raw data.
      col: The target database column name.

    Returns:
       - For 'object_value': A UTF-8 string truncated to 4096 bytes (safe-decoded).
        - For 'object_bytes': A Base64-encoded representation of the model's 'object_value'.
        - For other columns: The raw attribute value from the model.
    """
```

packages/datacommons-api/datacommons_api/services/graph_service.py
```diff
@@ -71,12 +79,22 @@ def create_node_model(graph_node: GraphNode) -> NodeModel:
         types = [types]
     types = [t for t in types if t is not None]
 
+    # Remove all CURIE namespaces before storing the node id
+    subject_id = strip_namespace(graph_node.id)
+    types = [strip_namespace(t) for t in types]
     return NodeModel(
```
clincoln8, line 85: long term when we revisit the code: it might be helpful to be more explicit for what this NodeModel is, maybe DbNodeModel? to differentiate from all of the other ways "Node" is used in datacommons.

packages/datacommons-api/datacommons_api/services/graph_service.py
```diff
@@ -203,6 +228,149 @@ def node_model_to_graph_node(node: NodeModel) -> GraphNode:
...
+        # A Spanner index key incorporates both the indexed columns AND the Primary Key.
+        # Max index key length is 8192 bytes total. The Primary Keys can swallow up to 4096 bytes easily.
+        # So we must restrict object_value to 4096 bytes to guarantee the total key size is < 8192 bytes.
+        if col == "object_value":
+            if len(val_bytes) > OBJECT_VALUE_MAX_LENGTH:
+                # Slice to exactly OBJECT_VALUE_MAX_LENGTH bytes, dropping fragmented chars gracefully
+                val_truncated = val_bytes[:OBJECT_VALUE_MAX_LENGTH].decode(
+                    "utf-8", errors="ignore"
+                )
+                return val_truncated
```
clincoln8: do we need to worry about hashing at all for this value? if we're just truncating, is there possibility for index key collision?

packages/datacommons-db/datacommons_db/models/edge.py
```diff
@@ -20,18 +20,23 @@
...
+    __tablename__ = EDGE_TABLE_NAME
     subject_id = sa.Column(
         String(1024), sa.ForeignKey("Node.subject_id"), primary_key=True
     )
     predicate = sa.Column(String(1024), primary_key=True)
     object_id = sa.Column(String(1024), primary_key=True)
-    object_value = sa.Column(Text(), nullable=True)
+    object_value = sa.Column(String(OBJECT_VALUE_MAX_LENGTH), nullable=True)
+    object_bytes = sa.Column(sa.LargeBinary(), nullable=True)
     object_hash = sa.Column(String(64), primary_key=True, nullable=True)
```
clincoln8, line 40: what is the object_hash used for again?

packages/datacommons-api/datacommons_api/services/graph_service.py
```diff
@@ -290,40 +465,60 @@ def _get_nodes_with_outgoing_edges(
...
+            len(node_models),
+            total_edges,
+            len(node_model_batches),
+        )
 
-        logger.info("Inserting %d nodes and %d edges", len(nodes), len(edges))
+        # Insert nodes and edges in batches
+        success_count = 0
+        try:
+            for node_model_batch in node_model_batches:
```
clincoln8, line 501: what happens if nodeA points to nodeB which is in a later batch and not in the graph yet. I think this might cause an error trying to write nodeA in the first batch because the edge pointing to nodeB would say that nodeB wasn't found in the node table yet?

If that seems plausible, we can mark as a todo to handle this case. Or ignore if that's not actually an issue.

packages/datacommons-api/datacommons_api/api_cli.py
```diff
@@ -72,3 +73,41 @@ def start(
...
+@click.option("--yes", is_flag=True, help="Skip confirmation prompt.")
+def drop_tables(
+    gcp_project_id: str,
+    gcp_spanner_instance_id: str,
+    gcp_spanner_database_name: str,
+    yes: bool,
+):
+    """Drop Node and Edge tables from the graph database."""
+    if not yes:
+        click.confirm(
```
clincoln8, line 96: just to check -- why are we dropping the tables? It seems like we'd want the user to be able to clear the table, but just dropping seems like it might be error prone. I think we do recreate the tables before writing if they're not already initialized, but still not exactly sure what scenario the user would want to drop them? It seems like that should be abstracted from a DCP owner.

clincoln8 (approved):
Thanks!!

packages/datacommons-api/datacommons_api/services/graph_service.py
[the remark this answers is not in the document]
dwnoble: I'll look into this as a follow up, but it appears to be writing the value properly to spanner 🤷

packages/datacommons-api/datacommons_api/services/graph_service.py
dwnoble: done

packages/datacommons-api/datacommons_api/services/graph_service.py
dwnoble: done

packages/datacommons-api/datacommons_api/services/graph_service.py
dwnoble: Good call- marking this as a todo for now.

this could apply to both nodes that are in later batches, and also nodes that are in remote knowledge graphs and not defined locally

packages/datacommons-api/datacommons_api/api_cli.py
dwnoble: I agree that dropping the data sounds like a better solution. I think i fussed with it and this ended up being easer in the moment, but we should revisit and just drop the data and keep the tables.

dwnoble: Added a todo

packages/datacommons-db/datacommons_db/models/edge.py
dwnoble: I am not sure- @n-h-diaz do you know what this field is for in the spanner db?

packages/datacommons-api/datacommons_api/services/graph_service.py
dwnoble: Good call- added a todo to look into this. Maybe we can truncate the value and append it to a hash of the entire byte array.

packages/datacommons-api/datacommons_api/services/graph_service.py
dwnoble: Agreed

PR #18, opened 2026-03-12
[DCP Deployment] Update Dockerfile for UV pinning & nonroot access

gmechali:
Also makes a small change to the cloudbuild to use the SHORT_SHA notation that github supports for the triggers to catch the git commit hash.


clincoln8 (approved):
Thank you!!

PR #19, opened 2026-03-12
[DCP] Use the image tag instead of latest in TF deployments

gmechali:
If you use the image tag "latest" , terraform won't know whether the latest image has changed and will not re-deploy the new image. This makes it more annoying for a customer who may want to update to the latest DCP image.

With the current changes, the dcp image tag is a variable passed in which defaults to latest, but if they run ./update_image_tag.sh, it will modify their terraform.tfvars file to apply the latest image's tag.
Therefore, they will be able to run terraform plan to view what would get re-deployed

### DIFF
--- a/infra/dcp/modules/cdc/locals.tf
+++ b/infra/dcp/modules/cdc/locals.tf
@@ -85,6 +85,15 @@ locals {
           version = "latest"
         }
       }
+    },
+    {
+      name = "MAPS_API_KEY"
+      value_source = {
+        secret_key_ref = {
+          secret  = var.disable_google_maps ? "" : google_secret_manager_secret.maps_api_key[0].secret_id
+          version = "latest"
+        }
+      }
     }
   ], var.use_spanner ? [] : [
     {
--- a/infra/dcp/modules/cdc/locals.tf
+++ b/infra/dcp/modules/cdc/locals.tf
@@ -53,10 +53,6 @@ locals {
       value = "gs://${local.gcs_data_bucket_name}/${var.gcs_data_bucket_output_folder}"
     },
     {
-      name  = "FORCE_RESTART"
-      value = "${timestamp()}"
-    },
-    {
       name  = "REDIS_HOST"
       value = try(local.redis_instance.host, "")
     },
--- a/infra/dcp/modules/cdc/locals.tf
+++ b/infra/dcp/modules/cdc/locals.tf
@@ -1,0 +1,90 @@
+# Local variable definitions
+
+locals {
+  # Data Commons Data Bucket
+  gcs_data_bucket_name = var.gcs_data_bucket_name != "" ? var.gcs_data_bucket_name : "${var.namespace}-datacommons-data-${var.project_id}"
+
+  # Use var.maps_api_key if set, otherwise use generated Maps API key
+  maps_api_key = var.maps_api_key != null ? var.maps_api_key : google_apikeys_key.maps_api_key[0].key_string
+
+  # Data Commons API hostname
+  dc_api_hostname = "api.datacommons.org"
+
+  # Data Commons API protocol
+  dc_api_protocol = "https"
+
+  # Data Commons API root URL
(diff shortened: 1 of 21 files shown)


dwnoble (changes requested):
@gmechali should i hold off on reviewing this PR until https://github.com/datacommonsorg/datacommons/pull/15 is submitted?

infra/dcp/update_image_tag.sh
```diff
@@ -0,0 +1,93 @@
...
+    fi
+fi
+
+# 3. Update terraform.tfvars
+echo "📝 Updating $TFVARS_FILE..."
+
+# Check if dcp_image_tag already exists
+if grep -q "^dcp_image_tag" "$TFVARS_FILE"; then
+    # Use sed to replace the line. (Special syntax for macOS)
+    sed -i '' "s/^dcp_image_tag.*/dcp_image_tag = \"$TAG\"/" "$TFVARS_FILE"
```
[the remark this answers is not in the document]
gmechali: resolved.

infra/dcp/modules/cdc/locals.tf
```diff
@@ -52,10 +52,6 @@ locals {
       name  = "OUTPUT_DIR"
       value = "gs://${local.gcs_data_bucket_name}/${var.gcs_data_bucket_output_folder}"
     },
-    {
-      name  = "FORCE_RESTART"
```
dwnoble, line 56: this env variable is a different approach to tackling the problem of re-deploying the "latest" or "stable" image.

Whenever a container env variable changes, cloud deploy will automatically re-deploy the image (thus grabbing the updated "latest" or "stable" image)

For now i would leave this in

infra/dcp/variables.tf
```diff
@@ -42,10 +42,16 @@ variable "enable_cdc" {
...
-variable "dcp_image_url" {
-  description = "Docker image URL for DCP"
+variable "dcp_image_base" {
+  description = "Docker image base URL for DCP (without tag)"
   type        = string
-  default     = "gcr.io/datcom-ci/datacommons-platform:latest"
+  default     = "us-docker.pkg.dev/datcom-ci/gcr.io/datacommons-platform"
+}
+
+variable "dcp_image_tag" {
```
dwnoble, line 51: I am OK separating out the image from the tag, but we should be consistent with all 3 images (datacommons-data, datacommons-services, and datacommons-platform)

infra/dcp/update_image_tag.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
```
dwnoble, line 1: Rather than tackling the problem with a script, I think we should consider adding semantic versioning to the 3x images published(datacommons-services, datacommons-data, and datacommons-platform).

For example, if we publish a new release with version `v1.2.3` on dcp from https://github.com/datacommonsorg/datacommons, that should trigger a new DCP image build with the tag `v1.2.3` (And maybe `stable` still?). The same should go for CDC releases. Today for CDC we publish a new image into the `customdc_stable` branch, but we should migrate this to publish the CDC images when our normal releases go out (like https://github.com/datacommonsorg/website/releases/tag/v3.4.4).

For the time being, i think adding the "FORCE_RESTART" env variable to all 3 images is more straightforward than the script to manually set the image version. what do you think @gmechali ?

dwnoble (changes requested):

infra/dcp/update_image_tag.sh
gmechali: So what I dont like about using the :latest and the FORCE_RESTART is that it becomes completely invisible to users.

If they run tf plan, they can't see whether it will deploy a new image. And just in general with FORCE_RESTART, they could deploy a new image without knowing.

This script gives them the warning, they run tf plan and see the exact image that's about to change. I think updating the image discretely is dangerous and clients should get the warning this provides.
WDYT?

PR #37, opened 2026-04-30
[DCP Ingestion] Connect the CDC Data Job to the ingestion workflow

gmechali:
Passes in the environment variables relating to the ingestion workflow into the CDC data job.
Ensures the SA running CDC data Job can trigger the workflow, and the SA running dataflow can write to the GCS bucket.


dwnoble (approved):

gmechali:
This PR is replaced by https://github.com/datacommonsorg/datacommons/pull/40

PR #50, opened 2026-05-16
Fix project resolution in admin ingest and improve TF operator transparency

clincoln8:
This PR addresses a bug where the `admin ingest start` command would infer the target GCP project from the user's ambient `gcloud` context instead of using the project explicitly configured in the Terraform state. It also removes the hardcoded region (`us-central1`) limitation by reading the region from Terraform outputs.

It also improves operator transparency and observability by parsing Cloud Run "operations" to provide direct, clickable links to the job in the GCP Console.

Finally, it resolves SSL/impersonation issues encountered when running the CLI tool in isolated environments (like `uvx` on macOS) by adding `pyopenssl` as a dependency.

#### CLI (`datacommons-admin`)
*   **Project & Region Resolution Fix**: `ingest start` now reads `project_id` and `region` from Terraform outputs instead of falling back to `google.auth.default()` or hardcoded defaults.
*   **Improved Operator Transparency**: Added support for parsing `operations` paths returned by the Cloud Run API. This allows the CLI to output direct, clickable links to the job in the GCP Console even for long-running async operations.
*   **Templates**: Updated the generated `main.tf` and `terraform.tfvars` templates to make the web service image configurable (defaulting to `stable`) and to expose `project_id`, `cdc_service_name`, and `region` outputs.

#### Terraform
*   **Outputs**: Exposed `project_id`, `cdc_service_name`, and `region` as outputs in the modules to support the CLI.

#### Dependencies
*   **`datacommons-admin`**: Added `pyopenssl>=24.0.0` to resolve SSL validation warnings in isolated environments.
*   **Root**: Declared the root project as a meta-package to prevent accidental code bundling.

## How to Test

You can test all these commands directly from this branch without installing them locally by using `uvx`.

**1. Generate Scaffolding:**
```bash
uvx --from "git+https://github.com/clincoln8/datacommons.git@uvx#subdirectory=packages/datacommons-cli" datacommons admin init
```

**2. Initialize Database:**
```bash
uvx --from "git+https://github.com/clincoln8/datacommons.git@uvx#subdirectory=packages/datacommons-cli" datacommons admin init-db
```

**3. Start Ingestion Job:**
```bash
uvx --from "git+https://github.com/clincoln8/datacommons.git@uvx#subdirectory=packages/datacommons-cli" datacommons admin ingest start
```


packages/datacommons-admin/datacommons_admin/ingestion_job_client.py
```diff
@@ -20,18 +20,20 @@
 class IngestionJobClient:
     """Client for interacting with Cloud Run Admin API to manage CDC data ingestion jobs."""
 
-    def __init__(self, job_name: str, service_account_email: str = None) -> None:
+    def __init__(self, job_name: str, service_account_email: str = None, project_id: str = None) -> None:
         self.service_account_email = service_account_email
-        base_credentials, project_id = google.auth.default()
+        base_credentials, default_project_id = google.auth.default()
+        
+        resolved_project_id = project_id or default_project_id
```
clincoln8: @dwnoble should we be falling back to the default project at all? or if it can't find the project_id from terraform then should it just error out?

packages/datacommons-admin/datacommons_admin/ingestion_job_client.py
```diff
@@ -20,18 +20,22 @@
...
+        base_credentials, default_project_id = google.auth.default()
+
+        resolved_project_id = project_id or default_project_id
 
         if not job_name.startswith("projects/"):
-            if not project_id:
+            if not resolved_project_id:
                 raise click.ClickException(
-                    "Could not determine GCP project ID from environment. Please verify your gcloud auth login."
+                    "Could not determine GCP project ID from environment or Terraform. Please verify your gcloud auth login."
```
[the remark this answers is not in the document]
clincoln8: done

packages/datacommons-admin/datacommons_admin/ingest_cli.py
```diff
@@ -77,6 +82,24 @@ def start() -> None:
             click.secho(job_url, fg="blue", underline=True)
             click.secho("Execution Console Link: ", fg="cyan", bold=True, nl=False)
             click.secho(exec_url, fg="blue", underline=True)
+        elif (
+            len(parts) >= 6
+            and parts[0] == "projects"
+            and parts[2] == "locations"
+            and parts[4] == "operations"
```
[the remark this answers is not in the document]
clincoln8: renamed to resolved_project_id

packages/datacommons-admin/datacommons_admin/ingestion_job_client.py
```diff
@@ -20,19 +20,25 @@
...
                 )
-            location = "us-central1"
-            self.full_job_name = (
-                f"projects/{project_id}/locations/{location}/jobs/{job_name}"
-            )
+            self.full_job_name = f"projects/{resolved_project_id}/locations/{resolved_location}/jobs/{job_name}"
```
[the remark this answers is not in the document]
clincoln8: @dwnoble lmk what you think about falling back to central for region. maybe address as a followup?

packages/datacommons-admin/datacommons_admin/ingest_cli.py
```diff
@@ -77,6 +88,24 @@ def start() -> None:
             click.secho(job_url, fg="blue", underline=True)
             click.secho("Execution Console Link: ", fg="cyan", bold=True, nl=False)
             click.secho(exec_url, fg="blue", underline=True)
+        elif (
```
gmechali, line 91: When would it go through here?

packages/datacommons-admin/datacommons_admin/ingestion_job_client.py
```diff
@@ -20,19 +20,25 @@
...
+        service_account_email: str = None,
+        project_id: str = None,
+        location: str = None,
+    ) -> None:
         self.service_account_email = service_account_email
-        base_credentials, project_id = google.auth.default()
+        base_credentials, default_project_id = google.auth.default()
+
+        resolved_project_id = project_id or default_project_id
+        resolved_location = location or "us-central1"
```
gmechali: The location is a requirement in terraform. We already set a default in the tf variable right? 
Can we avoid adding this here?

gmechali (approved):
Thanks Christie! Couple small questions but no objections

packages/datacommons-admin/datacommons_admin/ingestion_job_client.py
gmechali: > if it can't find the project_id from terraform
I think if this is the case, then something went seriously wrong. We shouldnt risk deploying something in a different project. I think we should throw an error if we somehow fail to extract the project ID.

packages/datacommons-admin/datacommons_admin/ingest_cli.py
clincoln8: I briefly looked into this and from my initial understanding [the endpoint](https://docs.cloud.google.com/run/docs/reference/rest/v2/projects.locations.jobs/run) actually only returns operations, not executions.

I might submit this as is and then remove the previous block after doing some more testing and also refactor the method a little bit.

packages/datacommons-admin/datacommons_admin/ingestion_job_client.py
clincoln8: changed to throwing exception if the location is not provided

packages/datacommons-admin/datacommons_admin/ingestion_job_client.py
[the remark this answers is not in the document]
clincoln8: throwing an error if location is not provided

packages/datacommons-admin/datacommons_admin/ingestion_job_client.py
clincoln8: agreed, changed this to throw an error if project_id is not provided via terraform

PR #51, opened 2026-05-18
[CLI + TF] Stop using copies of the Terraform main/variables/outputs terraform files

gmechali:
This switches us to using the version of those files checked in main branch. Otherwise we essentially needed a copy of those files, while we originally made slimmed down copies, it caused issues as all the variables are required, so it gave users only a limited subset of toggles.

Note that we override the source of the "stack" module to be github when you run the init command.


clincoln8 (approved):
Overall this seems like good progress, thanks!!

I'm still a little concerned about finer grained versioning. Like if someone first runs dcp-init and then after we publish changes tries to run ingestion-start, how do we keep things in sync? Will discuss offline.

gmechali:
> Overall this seems like good progress, thanks!!
> 
> I'm still a little concerned about finer grained versioning. Like if someone first runs dcp-init and then after we publish changes tries to run ingestion-start, how do we keep things in sync? Will discuss offline.

that's a good point but I do think we can make a solution. First of all, if we update the modules, then it's all safe and they get it immediately (which also means danger and we need versioning).
For the main/outputs/variables file, they would need to rerun the init script, which would override their tfvars, so we should definitely add an upgrade mode.

I'd say not for this week? but something to add to the list.

infra/dcp/terraform.tfvars.template
```diff
@@ -23,7 +23,7 @@ dcp_create_ingestion_bucket        = true
 dcp_ingestion_lock_timeout         = 82800
 
 # --- CDC ---
-# cdc_dc_api_key = "your-dc-api-key" # https://apikeys.datacommons.org
+cdc_dc_api_key = "$$DC_API_KEY$$" # https://apikeys.datacommons.org
```
clincoln8, line 26: nit: would it make sense to actually call this base_dc_api_key? just based on the variable name, it's a little confusing whether this api key is for the cdc instance or the remote/base instance

we can address in a follow up too

infra/dcp/terraform.tfvars.template
```diff
@@ -1,6 +1,6 @@
 # --- Shared Global Variables ---
-project_id = "your-project-id"
-namespace  = "your-namespace"
+project_id = "$$PROJECT_ID$$"
```
clincoln8, line 2: minor nit: should we rename this file to terraform.tfvars.**template** to convey how this file is used in practice?

packages/datacommons-admin/datacommons_admin/admin_cli.py
```diff
@@ -216,15 +237,46 @@ def init(
...
+        # Modify tfvars_example with actual values
+        tfvars_content = tfvars_example
+        tfvars_content = tfvars_content.replace(
+            '"$$PROJECT_ID$$"', f'"{resolved_project_id}"'
+        )
+        tfvars_content = tfvars_content.replace(
+            '"$$NAMESPACE$$"', f'"{resolved_namespace}"'
+        )
+        if resolved_dc_api_key:
+            tfvars_content = tfvars_content.replace(
```
clincoln8, line 267: if resolved_dc_api_key is empty (e.g., the user skipped the prompt), this if condition skips the replacement which would leave the literal token "$$DC_API_KEY$$" in the generated terraform.tfvars file

packages/datacommons-admin/datacommons_admin/admin_cli.py
```diff
@@ -216,15 +237,46 @@ def init(
...
-    )
+
+    try:
+        variables_content, main_content, outputs_content, tfvars_example = (
+            _get_github_templates(ref)
+        )
+
+        # Update the stack module source to point to GitHub
+        resolved_source = f"git::https://github.com/datacommonsorg/datacommons.git//infra/dcp/modules/stack?ref={ref}"
+        main_content = re.sub(
```
clincoln8, line 248: This regex replacement relies on the exact string source = "./modules/stack". If the format in infra/dcp/main.tf changes slightly (e.g., spacing), this will fail silently and leave the file pointing to a broken relative path. Consider adding a check to verify that the replacement actually occurred, and warning the user or failing fast if it didn't.

clincoln8 (approved):
An alternative might be to bundle the tf files directly into the module. There's some pros and cons for this, discussed here: https://docs.google.com/document/d/1_-E905HYxry2fLjQsZikKJOqxXqcZGbTLQIu3P7DuY8/edit?tab=t.0#heading=h.j4eaysunwjy3

infra/dcp/terraform.tfvars.template
gmechali: Good point - I ll leave this as part of the terraform module renaming (drop dcp/cdc) we ll need to do a lot of renaming.

infra/dcp/terraform.tfvars.template
gmechali: Fair, done.

packages/datacommons-admin/datacommons_admin/admin_cli.py
gmechali: I think that's fine, then they ll need to replace it manually.

packages/datacommons-admin/datacommons_admin/admin_cli.py
gmechali: I believe this is old, should support space differences but we also added a comment in the terraform.tfvars.example to say not to touch this because it's dangerous. So at lest if someone touches it they ll be alerted to test this.

gmechali:
> An alternative might be to bundle the tf files directly into the module. There's some pros and cons for this, discussed here: https://docs.google.com/document/d/1_-E905HYxry2fLjQsZikKJOqxXqcZGbTLQIu3P7DuY8/edit?tab=t.0#heading=h.j4eaysunwjy3

discussed in meeting - leaving as is.

PR #61, opened 2026-05-20
Adds the bq reservation resource

gmechali:
This PR fixes the creation of the Bigquery reservation, and the assignment. 
It also adds a variable to disable the bigquery reservation entirely since we just need one per GCP project, without the need to connect it to any job. 
And it adds variables for the slots.

### DIFF
--- a/infra/dcp/main.tf
+++ b/infra/dcp/main.tf
@@ -78,6 +78,8 @@ locals {
     spanner_database_id            = var.dcp_spanner_database_id
     spanner_version_retention_period = var.dcp_spanner_version_retention_period
     create_bq_reservation           = var.dcp_create_bq_reservation
+    bq_reservation_slot_capacity     = var.dcp_bq_reservation_slot_capacity
+    bq_reservation_max_slots        = var.dcp_bq_reservation_max_slots
     spanner_processing_units       = var.dcp_spanner_processing_units
     service_cpu                    = var.dcp_service_cpu
     service_memory                 = var.dcp_service_memory
--- a/infra/dcp/main.tf
+++ b/infra/dcp/main.tf
@@ -46,7 +46,8 @@ resource "google_project_service" "apis" {
     "dataflow.googleapis.com"
     ] : [], var.enable_bq_federation ? [
     "bigqueryconnection.googleapis.com",
-    "bigquery.googleapis.com"
+    "bigquery.googleapis.com",
+    "bigqueryreservation.googleapis.com"
   ] : []))
 
   service            = each.key
@@ -76,6 +77,7 @@ locals {
     spanner_instance_id            = var.dcp_spanner_instance_id
     spanner_database_id            = var.dcp_spanner_database_id
     spanner_version_retention_period = var.dcp_spanner_version_retention_period
+    create_bq_reservation           = var.create_bq_reservation
     spanner_processing_units       = var.dcp_spanner_processing_units
     service_cpu                    = var.dcp_service_cpu
     service_memory                 = var.dcp_service_memory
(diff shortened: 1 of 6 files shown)


infra/dcp/modules/spanner/bq_reservation.tf
```diff
@@ -0,0 +1,34 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Create the BigQuery Reservation for Federation queries
+resource "google_bigquery_reservation" "default" {
+  count         = var.create_spanner_db && var.enable_bq_federation ? 1 : 0
+  name          = "default"
```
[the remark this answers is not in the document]
gmechali: I will leave a default one for the time being - we can reevaluate this later.

infra/dcp/main.tf
```diff
@@ -88,8 +92,8 @@ locals {
     external_ingestion_bucket_name = var.dcp_external_ingestion_bucket_name
     ingestion_lock_timeout         = var.dcp_ingestion_lock_timeout
     ingestion_helper_image         = var.dcp_ingestion_helper_image
-    enable_bq_federation           = var.enable_bq_federation
-    bq_connection_name             = var.bq_connection_name
+    enable_bq_federation           = var.dcp_enable_bq_federation
```
dwnoble, line 95: i believe we're dropping the dcp_ prefix from variable names elsewhere. should we leave these as "enable_bq_federation" and "bq_connection_name" ?

gmechali: I was going to take a sledgehammer to that tomorrow, but once we start that, there's going to merge conflicts for anyone doing anything on terraform, so I wanted to make sure NO ONE had unsubmitted changes before starting.

infra/dcp/modules/spanner/variables.tf
```diff
@@ -64,3 +64,21 @@ variable "spanner_version_retention_period" {
...
+variable "create_bq_reservation" {
+  type        = bool
+  description = "Create a new BigQuery reservation for federation queries"
+  default     = true
+}
+
+variable "bq_reservation_slot_capacity" {
+  type        = number
+  description = "Baseline slots for BigQuery reservation"
+  default     = 100
```
dwnoble: @gmechali if we set the default to 0 does it still work if the autoscale max is 400? That way the cost goes to 0 if the bq federation isn't running

gmechali: great point - done.

infra/dcp/main.tf
dwnoble: okay sounds good. @clincoln8 i think may be taking a first pass at that tonight as well

infra/dcp/variables.tf
```diff
@@ -302,18 +302,36 @@ variable "cdc_search_scope" {
...
+variable "dcp_create_bq_reservation" {
+  description = "Create a new BigQuery reservation for federation queries"
+  type        = bool
+  default     = true
+}
+
+variable "dcp_bq_reservation_slot_capacity" {
+  description = "Baseline slots for BigQuery reservation"
+  type        = number
+  default     = 100
```
dwnoble: should this be 0 too?

dwnoble (approved):

infra/dcp/variables.tf
gmechali: ah yeah thanks!

PR #69, opened 2026-05-22
Fixes for Maps API Key naming conflicts and Spanner/Ingestion conditional guards

clincoln8:
- Added random suffix to Maps API key and secret to avoid recreation conflicts caused by GCP recovery state.
- Renamed Maps API key variables in the `auth` module for better consistency with root variables.
- Added `use_spanner` flag to `dataflow` and `helper_service` modules to prevent Spanner IAM bindings from being created when Spanner is disabled.
- Added a check block to ensure the Data Commons API key is provided when services or ingestion are enabled.
- Fixed the Spanner instance check block to not fail when Spanner is disabled.
- Clarified variable descriptions for Maps API key to explain precedence and destructive behavior.


infra/dcp/modules/stack/main.tf
```diff
@@ -149,24 +149,28 @@ module "ingestion_dataflow" {
...
+  namespace              = var.global.namespace
+  region                 = var.global.region
+  deletion_protection    = var.global.deletion_protection
   # Use index [0] because module.spanner is conditional. Fallback to empty string if disabled.
   spanner_instance_id    = var.spanner_config.enable ? module.spanner[0].spanner_instance_id : ""
   spanner_database_id    = var.spanner_config.enable ? module.spanner[0].spanner_database_id : ""
   bigquery_connection_id = var.spanner_config.enable ? module.spanner[0].bigquery_connection_id : ""
   ingestion_bucket_name  = module.storage.artifacts_bucket_name
   image                  = var.ingestion_config.helper_service_image
+  use_spanner            = var.spanner_config.enable
```
dwnoble, line 170: Is spanner required for our current setup?

dwnoble (approved):

infra/dcp/modules/stack/main.tf
clincoln8: pretty much. i'll rethink this and address in a followup

dwnoble: sounds good!

PR #71, opened 2026-05-22
[DCP - BigQuery Federation] Propagate the right variables for the BQ federtion to be authenticated, and run in the right region

gmechali:


infra/dcp/modules/ingestion/helper_service/main.tf
```diff
@@ -76,4 +76,29 @@ resource "google_storage_bucket_iam_member" "helper_bucket_access" {
...
+  role    = "roles/bigquery.dataEditor"
+  member  = "serviceAccount:${google_service_account.helper_sa[0].email}"
+}
+
+resource "google_project_iam_member" "helper_bq_job_user" {
+  count   = var.deploy ? 1 : 0
+  project = var.project_id
+  role    = "roles/bigquery.jobUser"
+  member  = "serviceAccount:${google_service_account.helper_sa[0].email}"
+}
```
clincoln8: nit: alternative option
```suggestion
resource "google_project_iam_member" "helper_bq_roles" {
  for_each = toset(var.deploy ? [
    "roles/bigquery.dataEditor",
    "roles/bigquery.jobUser"
  ] : [])
  project = var.project_id
  role    = each.value
  member  = "serviceAccount:${google_service_account.helper_sa[0].email}"
}
```

clincoln8: also do we want to condition on enable_postprocessing?

clincoln8 (approved):
Woohoo, thank you!!

infra/dcp/modules/ingestion/helper_service/main.tf
gmechali: yes and yes thanks

PR #72, opened 2026-05-22
Add ingestion_input_path to top level tf outputs

clincoln8:


gmechali (approved):

dhotchkiss (commented):
It looks like this has since been merged in a separate PR?

PR #76, opened 2026-05-24
[DCP - Ingestion] Clear the redis cache after each import

gmechali:
This is based on new PR which adds the clear_redis_cache action type to ingestion helper.
https://github.com/datacommonsorg/data/pull/2035

### DIFF
--- a/infra/dcp/modules/ingestion/workflow/main.tf
+++ b/infra/dcp/modules/ingestion/workflow/main.tf
@@ -110,15 +110,10 @@ resource "google_workflows_workflow" "ingestion_orchestrator" {
               - check_success:
                   switch:
                     - condition: '$${job_status.currentState == "JOB_STATE_DONE"}'
-                      next: check_bq_enabled
+                      next: %{if var.enable_bigquery_postprocessing}trigger_bq_federation%{else}promote_version%{endif}
               - fail_on_job_status:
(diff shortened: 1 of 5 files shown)


infra/dcp/modules/ingestion/workflow/main.tf
vish-cs, line 11: I see that we have the complete definition of the workflow added here. How do we ensure it stays in sync with the workflow definition for base DC? For example, I plan to make changes to the workflow to make aggregations asynchronous. How will that be incorporated here? We should avoid code duplication. Also, having the complete definition of a workflow embedded in a Terraform file does not seem a good idea to me.

gmechali: We do have the complete definition. For the record, the DCP workflow and the base DC one were supposed to have significant differences at the start since there was no aggregations needed for DCP etc. 
Now...well we had to come back on some of those and it's true the two are likely going to be the same at the end. 
Even the stuff we ve only added for DCP, I assume that's also needed for base.

Id love to find a way to share that, but as a follow up to these crazy timelines :)
Temporary duplication, but we do need to merge it back in.

vish-cs: Sure...let's push for a combined solution after the release deadline.

vish-cs (approved):

PR #82, opened 2026-05-26
Revert "[DCP - Services] - Fix mixer startup probe to prevent downtime & remove exponential backoff on lock acquisition"

gmechali:
Reverts datacommonsorg/datacommons#75

This is just the short term fix to buy us a few months to find the right long term fix. It will allow the deployment to succeed even as mixer and MCP continue to fail to startup properly before the init-db command is run

### DIFF
--- a/infra/dcp/modules/datacommons_services/main.tf
+++ b/infra/dcp/modules/datacommons_services/main.tf
@@ -45,11 +45,10 @@ resource "google_cloud_run_v2_service" "dc_web_service" {
       }
 
       startup_probe {
-        timeout_seconds   = 29
+        timeout_seconds   = 120
         period_seconds    = 30
-        failure_threshold = 30 # 30 * 30s = 15 minutes
-        http_get {
-          path = "/health"
+        failure_threshold = 6
+        tcp_socket {
           port = 8080
         }
       }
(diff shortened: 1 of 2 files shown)


infra/dcp/modules/datacommons_services/main.tf
```diff
@@ -45,11 +45,10 @@ resource "google_cloud_run_v2_service" "dc_web_service" {
       }
 
       startup_probe {
-        timeout_seconds   = 29
+        timeout_seconds   = 120
         period_seconds    = 30
```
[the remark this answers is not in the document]
keyurva: Can we address this?

infra/dcp/modules/datacommons_services/main.tf
```diff
@@ -45,11 +45,10 @@ resource "google_cloud_run_v2_service" "dc_web_service" {
...
-        failure_threshold = 30 # 30 * 30s = 15 minutes
-        http_get {
-          path = "/health"
+        failure_threshold = 6
```
keyurva: nit: Consider adding a comment what the value implies.

keyurva (approved):
Thanks!

infra/dcp/modules/datacommons_services/main.tf
gmechali: done

PR #87, opened 2026-05-27
[DCP - Ingestion] Updates to template for reservation support

gmechali:


infra/dcp/terraform.tfvars.template
```diff
@@ -51,10 +51,26 @@ auth_google_datacommons_api_key = "$$DC_API_KEY$$"
 # ingestion_input_path = "ingestion/input/"
 
 
+# =============================================================================
+# Ingestion - Postprocessing
+# =============================================================================
+# Set to true to create a dedicated BigQuery reservation for federation queries.
+#
```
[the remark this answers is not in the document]
gmechali: The defaults are correct, we no longer need these to be explicitly set

infra/dcp/terraform.tfvars.template
```diff
@@ -51,10 +51,26 @@ auth_google_datacommons_api_key = "$$DC_API_KEY$$"
...
+
 # =============================================================================
 # TEMP
 # =============================================================================
 auth_create_google_maps_api_key = false
-auth_google_maps_api_key = "TODO"
-spanner_enable_bigquery_connection                = true
-ingestion_workflow_enable_bigquery_postprocessing = false
\ No newline at end of file
+auth_google_maps_api_key = "TODO"
```
kmoscoe, line 72: Why not get rid of this and just "your-maps-api-key" as you do with the other variables above?

kmoscoe (commented):
General comment: Could we please remove the # from all variables that have a default setting? Otherwise it's very easy to overlook removing it and screwing up your deployment several times.

infra/dcp/terraform.tfvars.template
```diff
@@ -51,10 +51,26 @@ auth_google_datacommons_api_key = "$$DC_API_KEY$$"
 # ingestion_input_path = "ingestion/input/"
 
 
+# =============================================================================
+# Ingestion - Postprocessing
+# =============================================================================
+# Set to true to create a dedicated BigQuery reservation for federation queries.
```
kmoscoe: Rewrite this as follows:
This is needed for running BigQuery federated queries against the Spanner data. Set this to true for ONLY ONE deployment to create a dedicated reservation. Set to false for all other deployments in the same region and project to share it. Once you have created the reservation, do the following to remove it from Terraform state but preserve the reservation:
1. Run terraform state rm module.stack.module.spanner[0].google_bigquery_reservation.default[0]
terraform state rm module.stack.module.spanner[0].google_bigquery_reservation_assignment.project_assignment[0]
2. Set spanner_create_bigquery_reservation = false here.

infra/dcp/terraform.tfvars.template
```diff
@@ -51,10 +51,26 @@ auth_google_datacommons_api_key = "$$DC_API_KEY$$"
...
+# 
+# NOTE: BigQuery reservations are project-wide per region. If managing multiple 
+# deployments in the SAME project & region:
+# - Set this to true for ONLY ONE deployment to create the reservation.
+# - Set this to false for all other deployments to share it.
+# 
+# CAUTION: Deleting this reservation will break post-processing for ALL deployments
```
kmoscoe: What do you mean by "deleting"? I assume you don't mean setting it to false, but deleting the line altogether?

infra/dcp/terraform.tfvars.template
gmechali: Cuz im not touching this in this PR. this is for another PR later!

kmoscoe (approved):

infra/dcp/terraform.tfvars.template
gmechali: I meant destroying the resource.
Which if they dont remove it from the state is something that can happen. Note this is only an issue for multiple deployments per region/project

PR #94, opened 2026-05-29
Updated admin init command to fetch modules with the current version ref / tag

dwnoble:
This pull request updates the default behavior of the `admin init` command in the Data Commons Admin CLI to use the current package version (prefixed with "v") as the default Git ref for module sources, instead of always using "main". It also adds a test to ensure this behavior.

**Default Git ref update:**

* The `--ref` option in the `init` command now defaults to `v{__version__}` (e.g., "v1.2.3") instead of "main", ensuring that the generated infrastructure uses the module version matching the CLI version. [[1]](diffhunk://#diff-8f0ad2d713ed83ed888ce09972ad56fe0091b8ece926508adb0f9be0dc192c7bR25) [[2]](diffhunk://#diff-8f0ad2d713ed83ed888ce09972ad56fe0091b8ece926508adb0f9be0dc192c7bL384-R385)

**Testing:**

* Added a test (`test_init_uses_default_ref_v_prefixed`) to verify that the `init` command uses the version-prefixed default ref when none is specified.


packages/datacommons-admin/tests/test_admin_cli.py
```diff
@@ -376,3 +376,33 @@ def test_ingest_show_config_success(
...
+
+@patch("datacommons_admin.admin_cli._get_github_templates")
+def test_init_uses_default_ref_v_prefixed(
+    mock_get_templates, runner: CliRunner, tmp_path: Path
+) -> None:
+    mock_get_templates.return_value = (
+        'variable "test" {}',
+        'module "stack" {\n  source = "./modules/stack"\n}',
+        'output "test" {}',
+        'project_id = "$$PROJECT_ID$$"\nnamespace  = "$$NAMESPACE$$"\n# cdc_dc_api_key = "$$DC_API_KEY$$"',
```
gmechali, line 389: this is the old name right? Can we update it

gmechali (approved):

packages/datacommons-admin/datacommons_admin/admin_cli.py
```diff
@@ -381,7 +382,7 @@ def _setup_dcp_config_dir(
 )
 @click.option("--dc-api-key", default="", help="Data Commons API key.")
 @click.option(
-    "--ref", default="main", show_default=True, help="Git ref for module source."
+    "--ref", default=f"v{__version__}", show_default=True, help="Git ref for module source."
```
clincoln8: nit: can this flag name be more explicit like `tf-git-ref` or something?

packages/datacommons-admin/tests/test_admin_cli.py
dwnoble: done

packages/datacommons-admin/datacommons_admin/admin_cli.py
dwnoble: done

PR #101, opened 2026-06-03
Uncomment variables from template file

kmoscoe:
Having necessary variables commented out makes it easy to overlook them and screw up the deployment. (See for example https://buganizer.corp.google.com/issues/503497509#comment7.)

This PR also:
- Sets redis to false, because this does not need to be created during development.
- Removes option to create Spanner database -- this is always needed, and removing it also reduces risk of erroneously setting it to false.
- Adds Analytics ID to template
- Removes Spanner DB name 
- Adds comments to all variables, thanks to @dwnoble


infra/dcp/terraform.tfvars.template
```diff
@@ -8,47 +8,49 @@
...
 project_id = "$$PROJECT_ID$$"
 namespace  = "$$NAMESPACE$$"
-# region     = "us-central1"
+region     = "us-central1"
 # Protects resources that hold persistent data (Spanner, GCS). Enable in prod!
-# stateful_deletion_protection = false
+stateful_deletion_protection = false
 # Protects compute resources (Cloud Run, Workflows). Keep false for easy updates.
-# stateless_deletion_protection = false
+stateless_deletion_protection = false
```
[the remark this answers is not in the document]
kmoscoe: Agree.

infra/dcp/terraform.tfvars.template
```diff
@@ -8,47 +8,49 @@
...
 # Get a free DC API Key at https://apikeys.datacommons.org (Required)
 auth_google_datacommons_api_key = "$$DC_API_KEY$$"
+auth_create_google_maps_api_key = false
+auth_google_maps_api_key = "your-maps-api-key"
 
 # =============================================================================
 # Storage Module
 # =============================================================================
-# storage_create_artifacts_bucket = true
+storage_create_artifacts_bucket = true
```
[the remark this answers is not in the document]
kmoscoe: Done.

infra/dcp/terraform.tfvars.template
```diff
@@ -8,47 +8,49 @@
...
 # storage_artifacts_bucket_name   = "your-artifacts-bucket-name"
 
 # =============================================================================
 # Redis Module
 # =============================================================================
-# enable_redis = true
+enable_redis = false
```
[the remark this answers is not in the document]
kmoscoe: done

infra/dcp/terraform.tfvars.template
```diff
@@ -8,47 +8,49 @@
...
 
 # =============================================================================
 # Spanner Module
 # =============================================================================
-# spanner_create_instance = true
+spanner_create_instance = true
 # spanner_instance_id     = "your-existing-spanner-instance"
-# spanner_create_database = true
-# spanner_database_id     = "dc-db"
+spanner_create_database = true
```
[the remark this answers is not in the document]
kmoscoe: done

infra/dcp/terraform.tfvars.template
```diff
@@ -8,47 +8,49 @@
...
 # Datacommons Services Module
 # =============================================================================
-# datacommons_services_allow_unauthenticated_access = true
+datacommons_services_allow_unauthenticated_access = true
 
 # =============================================================================
 # Ingestion - Preprocessing Job
 # =============================================================================
-# ingestion_input_path = "ingestion/input"
+ingestion_input_path = "ingestion/input"
```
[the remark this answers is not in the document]
kmoscoe: done

infra/dcp/terraform.tfvars.template
```diff
@@ -29,28 +31,28 @@ auth_google_datacommons_api_key = "$$DC_API_KEY$$"
 # =============================================================================
 # Redis Module
 # =============================================================================
-# enable_redis = true
+# enable_redis = false
```
gmechali: So the default of enable_redis is already false, so what's the point of this change here?

enable_redis is commented out here. I guess the point of having it true is if they uncomment it, that means they want it. Now if they uncomment it, it has 0 effect, because it would remain false.

infra/dcp/terraform.tfvars.template
```diff
@@ -29,28 +31,28 @@ auth_google_datacommons_api_key = "$$DC_API_KEY$$"
...
 # redis_instance_name = "dc-redis-instance"
 
 # =============================================================================
 # Spanner Module
 # =============================================================================
-# spanner_create_instance = true
+spanner_create_instance = true
```
gmechali, line 50: I dont necessarily disagree with this change, but we intentionally commented it out because there is some not insignificant cost to starting the instance. I think it's at least worth a comment saying you can set this to false and pass in a spanner_instance_id if you already have one you want to reuse to save $$$

infra/dcp/terraform.tfvars.template
```diff
@@ -29,28 +31,28 @@ auth_google_datacommons_api_key = "$$DC_API_KEY$$"
...
 # spanner_instance_id     = "your-existing-spanner-instance"
 # spanner_create_database = true
-# spanner_database_id     = "dc-db"
+spanner_database_id     = "dc-db"
```
gmechali: Same as with redis. The default variable is dc-db, so uncommenting it does nothing. The point was if they want to change the name, they can uncomment and change the name.

infra/dcp/terraform.tfvars.template
```diff
@@ -29,28 +31,28 @@ auth_google_datacommons_api_key = "$$DC_API_KEY$$"
...
 
 # =============================================================================
 # Datacommons Services Module
 # =============================================================================
-# datacommons_services_allow_unauthenticated_access = true
+datacommons_services_allow_unauthenticated_access = true
```
gmechali: Idk... I might argue leaving private until they explicitly make it public...

This is their data they're loading. Making the service public should be a decision, not a mistake.

infra/dcp/terraform.tfvars.template
```diff
@@ -62,5 +64,4 @@ spanner_create_bigquery_reservation = true
 # =============================================================================
```
gmechali: Yo ucan remove this entire temp section now then?

infra/dcp/terraform.tfvars.template
```diff
@@ -19,6 +19,8 @@ namespace  = "$$NAMESPACE$$"
...
 # Get a free DC API Key at https://apikeys.datacommons.org (Required)
 auth_google_datacommons_api_key = "$$DC_API_KEY$$"
+# auth_create_google_maps_api_key = true
```
gmechali: @dwnoble  wdyt about this?
 can we just remove it from the .template file?

infra/dcp/terraform.tfvars.template
kmoscoe: That's true. However, apparently, according to Gemini's review, Terraform wants variables that are set to the default to be commented out. The idea here is just to highlight it. Same as, for example, the region. WDYT?

infra/dcp/terraform.tfvars.template
kmoscoe: Yeah, I think it's unlikely that they will have an existing Spanner instance, so this is kind of annoying to have to uncomment. I will add your suggestion.

infra/dcp/terraform.tfvars.template
kmoscoe: Yep, good point. Fixed.

kmoscoe: Actually, no, in variables.tf it is set to empty string, so it's not obvious that the default is set to dc-db (notwithstanding the comment). In fact, this variable is one on my list that is confusing; it's not clear that if you set it it will completely override everything, including the namespace, or not. I actually wanted to verify the behavior of this and the storage-bucket-name; their behavior and name should be consistent, but I'm not sure what they are; I was planning to do this in a follow-up PR.

infra/dcp/terraform.tfvars.template
kmoscoe: OK, what it should be then, is set to false here, so it is very obvious what's going to happen. Users shouldn't have to look up in variables.tf what the default is. The main problem I have here is setting the default in the template to different from what it is in the variables.tf file. It's somewhat confusing here. I understand doing it for the bigquery reservation, but here, if you really feel it should be false by default, let's leave it as such.

infra/dcp/terraform.tfvars.template
kmoscoe: Yep, thanks!

infra/dcp/terraform.tfvars.template
kmoscoe: I think it's important to keep it. Because it's set by default to true, so if we keep it out of here, they have to add it back in.

infra/dcp/terraform.tfvars.template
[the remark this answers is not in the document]
kmoscoe: Actually decided to remove it entirely. No good  reason to have it in the template.

infra/dcp/terraform.tfvars.template
gmechali: https://github.com/datacommonsorg/datacommons/blob/17482f99f36188c94110aca9ee757ad22db36217/infra/dcp/modules/spanner/main.tf#L4
This is where we default the name to be dc-db, unless you specify a name.

infra/dcp/terraform.tfvars.template
gmechali: Terraform doesnt care about what's commented out. It doesnt read it. 

I think it's more clear to the user to have enable_redis = true as commented, and when they uncomment it, it enables redis, but that's just my opinion..

infra/dcp/terraform.tfvars.template
gmechali: That's also fine, but they shoud also understand these are commented out.
The comments can also be interpreted as suggestions of likely changes they need to implement to this file.

Keep in mind, they can add every single terraform variable, so this is in no way an exhaustive definition of our setup. It just highlights ones we think they ll need/want to change.

infra/dcp/terraform.tfvars.template
gmechali: They have to add it back in, if they care about disabling it. 

If they disable it, they lose core functionality. I think this is a very unlikely scenario for our users. So that's why Im hesitant on whehter it needs to be in the .template file. But if we want to call out the Maps api usage, then fine to keep

infra/dcp/terraform.tfvars.template
kmoscoe: Sorry, I meant Terraform best practices, according to Gemini Code-Assist. It states -- and I agree with it -- that if the value in terraform.tfvars is the same as a default set elsewhere, it should be commented out. Still disagree with you about the Redis default. :-)  Furthermore, it should be renamed to redis_create_instance, but that's for another PR.

infra/dcp/terraform.tfvars.template
kmoscoe: Yes, I know. But the users don't see that file. I mean yeah, they can look under the .terraform hood that gets created by datacommons admin init, but they would have to dig for it.
Also, just to mention -- storage_bucket_name and spanner_database_id use exactly opposite terminology: for GCS, the name replaces the entire string, including the prefix and suffix; for Spanner, it only replaces the unprefixed string -- this is another thing I would like to fix...

infra/dcp/terraform.tfvars.template
kmoscoe: I agree that it's unlikely. But it's also somewhat confusing if they already have their own key. Then they might wonder why there are 2 keys. This way, they can see that we create it by default.

kmoscoe: Sorry, what I meant to say is that by having these default listed in this file, it acts as a kind of documentation. Yes, they can see it in variables.tf, but I have a strong feeling most users won't look at that. So this way, if they really want to disable it, it's easy.

infra/dcp/terraform.tfvars.template
kmoscoe: Also, in Spanner itself, the terminology is "name",  not ID, so we should fix that anyway.

infra/dcp/terraform.tfvars.template
dwnoble: i think we should remove the google maps configuration from the variables file. i think it would be pretty rare for a customer to ever want to define their own key vs letting terraform manage the key creation.

infra/dcp/terraform.tfvars.template
```diff
@@ -29,38 +31,33 @@ auth_google_datacommons_api_key = "$$DC_API_KEY$$"
 # =============================================================================
 # Redis Module
 # =============================================================================
-# enable_redis = true
+# enable_redis = false
 # redis_instance_name = "dc-redis-instance"
```
dwnoble: Can we remove the redis instance name?

infra/dcp/terraform.tfvars.template
dwnoble: Gemini told me the same thing it told Kara, that it's best practice to show commented out variables set to their default value. However, gemini code assist says the opposite, so now I don't know who to believe! I don't really care which way we go, as long as we're consistent : https://screenshot.googleplex.com/9ykbCUwvVshcdh6

infra/dcp/terraform.tfvars.template
dwnoble: I think we should leave out both spanner_database_id (we can change it to `spanner_database_name` in a follow up) and redis name. users shouldn't need to customize the resource names

infra/dcp/terraform.tfvars.template
kmoscoe: Yes, as long as setting up GCP is the first step in their onboarding journey.  I'm _hoping_ that by "public preview" we  would have a way for them to run everything locally, in which case they would need the key beforehand. So I think it's worth keeping here.

kmoscoe: But if you feel strongly, I can take it out.

infra/dcp/terraform.tfvars.template
kmoscoe: It was actually Gemini Code Assist that told me this in several places!

infra/dcp/terraform.tfvars.template
kmoscoe: Sure, I don't see any reason why they need to set this.

kmoscoe: Done.

infra/dcp/terraform.tfvars.template
kmoscoe: SGTM, done.

infra/dcp/terraform.tfvars.template
kmoscoe: Here's my proposal: Let's uncomment stuff that has to be set in the first deployment, so it's clear to users that it's being applied -- even if the default is the same as in variables.tf. (Since Gemini Code Assist seems to be contradictory on this point!) And we can keep commented optional variables that they might want to set.

infra/dcp/terraform.tfvars.template
kmoscoe: OK took it out to simplify things.

infra/dcp/terraform.tfvars.template
```diff
@@ -8,59 +8,74 @@
...
 namespace  = "$$NAMESPACE$$"
-# region     = "us-central1"
-# Protects resources that hold persistent data (Spanner, GCS). Enable in prod!
-# stateful_deletion_protection = false
-# Protects compute resources (Cloud Run, Workflows). Keep false for easy updates.
-# stateless_deletion_protection = false
+region     = "us-central1"
+
+# Deletion protection for stateful resources holding persistent data 
+# (e.g., Cloud Spanner, Cloud Storage).
```
gmechali: It's just cloud spanner and storage. That is the full list.

The stateless, cloud run and workflows is not the full list.

infra/dcp/terraform.tfvars.template
```diff
@@ -8,59 +8,74 @@
...
+# spanner_instance_id     = "existing-spanner-instance"
 
 # =============================================================================
 # Datacommons Services Module
 # =============================================================================
-# datacommons_services_allow_unauthenticated_access = true
+# Controls whether the Data Commons API and frontend services are publicly accessible.
+# Set to true to allow unauthenticated public internet traffic to access the service.
+# Set to false (default) to restrict access (e.g., behind authentication, API 
+# gateways, or Cloud Identity-Aware Proxy).
```
gmechali: It's just behind Cloud IAM proxy I believe. 
They could setup their own api gateway, even configure apigee or whatever. but as is, this just requires IAM auth

infra/dcp/terraform.tfvars.template
```diff
@@ -8,59 +8,74 @@
...
 # =============================================================================
-# Note: You must have only one capacity reservation per project per region. 
-# If you intend to deploy multiple environments in the same project, please
-# carefully read the documentation.
+# Provisions a dedicated BigQuery reservation (slot commitment) for Spanner federated 
+# queries (used internally to boost performance). This ensures query capacity for
+# postprocessing / ingestion steps.
+# NOTE: Only one reservation can exist per GCP project per region. If you are deploying
+# multiple environments (e.g. dev and prod) in the same project, set this to
+# false for all other deployments.
```
gmechali: "set this to true for one deployment, and false for all others. See docs for more details.".

gmechali (approved):

infra/dcp/terraform.tfvars.template
kmoscoe: fixed.

infra/dcp/terraform.tfvars.template
kmoscoe: OK, so let's just simplify it and say what I said in the doc and point to that.

infra/dcp/terraform.tfvars.template
kmoscoe: Done.

PR #111, opened 2026-06-07
update workflow and admin client to use RESTful ingestion-helper endpoints

dwnoble:
## Description
Updates the Spanner ingestion orchestration workflow and the administrative Python client to align with the new RESTful FastAPI endpoints introduced in the updated `ingestion-helper` service.

## Changes
* **Orchestration Workflow ([main.tf](file:///Users/dwnoble/Projects/datacommons/datacommons/infra/dcp/modules/ingestion/workflow/main.tf))**: Refactored HTTP POST steps to call specific RESTful endpoints (e.g., `/database/lock/acquire`, `/imports/status`, `/aggregation/run`, etc.) and removed the legacy `actionType` request body parameter.
* **Admin Client ([ingestion_helper_client.py](file:///Users/dwnoble/Projects/datacommons/datacommons/packages/datacommons-admin/datacommons_admin/ingestion_helper_client.py))**: Refactored `IngestionHelperClient` to call `/database/initialize` and `/database/seed` directly and updated error response parsing to handle standard FastAPI `detail` payloads.

## Coordinated Merge Required
This PR is paired with the helper service changes in [datacommonsorg/import#526](https://github.com/datacommonsorg/import/pull/526/). **Both PRs must be merged at the same time** to avoid breaking the ingestion pipeline and administrative database setup commands.

### DIFF
--- a/infra/dcp/modules/ingestion/workflow/main.tf
+++ b/infra/dcp/modules/ingestion/workflow/main.tf
@@ -52,11 +52,10 @@ resource "google_workflows_workflow" "ingestion_orchestrator" {
           try:
             call: http.post
             args:
-              url: '${var.ingestion_helper_url}'
+              url: '${var.ingestion_helper_url}/database/lock/acquire'
               auth:
                 type: OIDC
               body:
-                actionType: "acquire_ingestion_lock"
                 workflowId: '$${workflow_id}'
                 timeout: '$${lock_timeout}'
             result: lock_result
@@ -73,11 +72,10 @@ resource "google_workflows_workflow" "ingestion_orchestrator" {
               - set_import_staging:
                   call: http.post
                   args:
-                    url: '${var.ingestion_helper_url}'
+                    url: '${var.ingestion_helper_url}/imports/status'
                     auth:
                       type: OIDC
                     body:
-                      actionType: "update_import_status"
                       importName: '$${input.importName}'
                       status: "STAGING"
                       latestVersion: '$${latest_version_gcs_path}'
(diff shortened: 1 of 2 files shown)


infra/dcp/modules/ingestion/workflow/main.tf
```diff
@@ -52,11 +52,10 @@ resource "google_workflows_workflow" "ingestion_orchestrator" {
           try:
             call: http.post
             args:
-              url: '${var.ingestion_helper_url}'
+              url: '${var.ingestion_helper_url}/database/lock/acquire'
```
gmechali, line 55: I think the way you made those restful endpoints is very nested. 

Realistically every restful resource will end up with a single  endpoint. Wdyt about jsut database/lock (with param acquire/release)
database/status (with parameter = staging)
database/version

then maybe
postprocessing/run_aggregations
postprocessing/generate_embeddings ?

(not tied to these at all but something like it - non blocking comment, approving)

gmechali (approved):

infra/dcp/modules/ingestion/workflow/main.tf
```diff
@@ -157,12 +154,11 @@ resource "google_workflows_workflow" "ingestion_orchestrator" {
...
                                 call: http.post
                                 args:
-                                  url: '${var.ingestion_helper_url}'
+                                  url: '${var.ingestion_helper_url}/embeddings/ingest'
                                   auth:
                                     type: OIDC
                                   body:
-                                    actionType: "embedding_ingestion"
-                                    importList: '$${json.decode(input.importList)}'
+                                    enableEmbeddings: true
```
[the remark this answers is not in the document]
dwnoble: removing importName looks correct

infra/dcp/modules/ingestion/workflow/main.tf
dwnoble: jetski had some hesitations about combining the endpoints together - ill keep it like this for now, but we can revisit this offline and come to a consensus

PR #119, opened 2026-06-12
add quick validation for the api key and add log if api key isn't valid

yiyche:
Add API key validations:
1. Local config parser (get_tfvars_api_key): Added a parser in tf_utils.py to extract the API key from the local terraform.tfvars 
2. Triggers a quick validation request against api.datacommons.org/v2/node. If the server returns an HTTP 401 Unauthorized, the CLI prints a clear warning and aborts execution immediately


dwnoble (commented):
I like doing a check for the API key before starting the ingestion job, but I think the logistics of it should go in a different direction.

TLDR; instead of doing the api key check in the CLI, can you add a new new ingestion helper endpoint (maybe as a new "ingestion.py" router here: https://github.com/datacommonsorg/import/tree/master/pipeline/workflow/ingestion-helper/routes) that does the api key check and also starts the cloud run job? then update the datacommons admin ingest start command to call the ingestion helper api endpoint instead of starting cloud run directly

Rationale:
Today, we start the ingestion job by invoking a cloud run job directly, and on completion that job starts a cloud workflow. Eventually we want to move the cloud run workflow. From a user perspective, they shouldn't care whether we start a cloud run job or workflow, they should just know that ingestion has started and how they can check on it.


Also, in addition to checking if the API key exists, we should also check things like:
- does the input data artifact gcs bucket exist?
- is there a config.json and other required files in the input directory in the above bucket?
- is there an ingestion job already running?
- probably more down the road

last thing- here the CLI checks if the api key in the terraform.tfvars is valid, but it really should be checking if the actual secret in GCP is valid (example: https://pantheon.corp.google.com/security/secret-manager/secret/itest-4771-dc-api-key/versions?e=13803378&mods=-monitoring_api_staging&project=datcom-ci). while this is totally possible, i think it's more elegant to instead update terraform to set a DC api key in the environment for the ingestion helper, then just have the ingestion helper check if its local key is valid

packages/datacommons-admin/datacommons_admin/tf_utils.py
```diff
@@ -134,3 +132,33 @@ def get_region() -> str:
 def get_ingestion_workflow_name() -> str:
     """Convenience wrapper to fetch the ingestion_workflow_name Terraform output."""
     return get_terraform_output(TF_OUTPUT_INGESTION_WORKFLOW_NAME)
+
+
+def get_tfvars_api_key() -> str | None:
```
dwnoble: Can this helper be removed now?

packages/datacommons-admin/datacommons_admin/ingest_cli.py
```diff
@@ -40,36 +43,41 @@ def ingest() -> None:
...
     click.secho(f"Found Workflow Service Account: {sa_email}", fg="green")
     click.secho(f"Found GCP Project ID: {project_id}", fg="green")
     click.secho(f"Found GCP Region: {region}", fg="green")
     click.secho(
-        f"Starting Cloud Run job '{job_name}' via Admin API (this may take a few moments)...",
+        "Triggering data ingestion via Ingestion Helper API...",
         fg="bright_black",
     )
 
-    client = IngestionJobClient(
```
dwnoble, line 63: Once the /ingestion/start endpoint is in place, the IngestionJobClient can be removed entirely

packages/datacommons-admin/datacommons_admin/ingest_cli.py
```diff
@@ -40,36 +43,41 @@ def ingest() -> None:
...
-        job_name,
+    client = IngestionHelperClient(
+        helper_url,
         service_account_email=sa_email,
-        project_id=project_id,
-        location=region,
     )
-    result = client.start_job(imports=imports)
+    result = client.start_ingestion(job_name, imports=imports)
```
dwnoble: is it necessary to pass the job_name to the ingestion start endpoint?

dwnoble (commented):
Thanks @yiyche - have you started migrating the IngestionJobClient to the ingestion helper?

yiyche:
> Thanks @yiyche - have you started migrating the IngestionJobClient to the ingestion helper?

just did - PTAL, thank you!

kmoscoe:
> I like doing a check for the API key before starting the ingestion job, but I think the logistics of it should go in a different direction.
> 
> TLDR; instead of doing the api key check in the CLI, can you add a new new ingestion helper endpoint (maybe as a new "ingestion.py" router here: https://github.com/datacommonsorg/import/tree/master/pipeline/workflow/ingestion-helper/routes) that does the api key check and also starts the cloud run job? then update the datacommons admin ingest start command to call the ingestion helper api endpoint instead of starting cloud run directly
> 
> Rationale: Today, we start the ingestion job by invoking a cloud run job directly, and on completion that job starts a cloud workflow. Eventually we want to move the cloud run workflow. From a user perspective, they shouldn't care whether we start a cloud run job or workflow, they should just know that ingestion has started and how they can check on it.
> 
> Also, in addition to checking if the API key exists, we should also check things like:
> 
> * does the input data artifact gcs bucket exist?
> * is there a config.json and other required files in the input directory in the above bucket?
> * is there an ingestion job already running?
> * probably more down the road
> 
> last thing- here the CLI checks if the api key in the terraform.tfvars is valid, but it really should be checking if the actual secret in GCP is valid (example: https://pantheon.corp.google.com/security/secret-manager/secret/itest-4771-dc-api-key/versions?e=13803378&mods=-monitoring_api_staging&project=datcom-ci). while this is totally possible, i think it's more elegant to instead update terraform to set a DC api key in the environment for the ingestion helper, then just have the ingestion helper check if its local key is valid

Hey both --
I also like the idea of doing it in the CLI or in the workflow. I think it may also need to be done in Terraform. I could see many cases where they only want to do something like update an environment variable or change some other parameters on the services container, without running any ingestion workflow.

PR #122, opened 2026-06-17
turn on resolve with spanner embedding to default to true

shixiao-coder:


infra/dcp/variables.tf
```diff
@@ -306,7 +306,7 @@ variable "datacommons_services_mcp_instructions_path" {
 variable "datacommons_services_resolve_with_spanner_embeddings" {
```
gmechali, line 307: nit: Can you add a TODO to later completely remove this variable? 

Fine to keep for now but if i understand correctly, we are essentially ready to make Spanenr embeddings the only embeddings, and so we can soon just make this entirely dependent on the terraform variable for creating the spanner embeddings!

gmechali (approved):
Thanks Xiao! Leaving as is for now is fine! we should get it completely removed long term. but i think it's fair to do that later once the M0 milestone is no longer in use

infra/dcp/variables.tf
```diff
@@ -306,7 +306,7 @@ variable "datacommons_services_mcp_instructions_path" {
...
   description = "Enable resolving search queries with Spanner embeddings"
   type        = bool
-  default     = false
+  default     = true
```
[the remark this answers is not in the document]
shixiao-coder: updated

infra/dcp/variables.tf
shixiao-coder: sure, added the TODO

PR #130, opened 2026-06-18
Fix Details: None output in admin cli database commands

clincoln8:
Fixes a minor display bug where 'Details: None' was printed upon successful database initialization or seeding. The CLI was checking for key existence ('message' in result) instead of checking if the message was truthy, leading to printing None when the API returned null.


shixiao-coder (approved):
LGTM

PR #132, opened 2026-06-18
create a sandbox integration test on GCP

yiyche:
roughly this test 
- Spins up real sandboxed GCP infrastructure.
- Deploys updated multi-entity Spanner schemas.
- checks GCS asset uploads and Dataflow runs.
- Executes Cloud Workflows database aggregation jobs
- Verifies raw nodes are queryable in Spanner
- Tests NL search
- Validates container network proxies end-to-end
- Tears down at the end

This test is expensive to run, so this should be run only once per day / when we try to create candidate for release. **the PR to configure this will come as a follow up once infra is more stable**

To run the test: `uv run python3 tests/datacommons-integration-tests/run_gcp_integration_test.py --project-id <project_id>`

amending: all tests will reuse `dcp-integration-test-shared-instance` spanner instance since creating new instance is expensive and most likely not something we will end up needing to test.

### DIFF
--- a/tests/datacommons-integration-tests/README.md
+++ b/tests/datacommons-integration-tests/README.md
@@ -48,9 +48,53 @@ While running with `--keep-containers`, you can access:
 
 ## Custom Feature Flags
 
-Local integration tests run with the Spanner multi-entity schema enabled by default in the Mixer backend. This is controlled by mounting [custom_feature_flags.yaml](file:///usr/local/google/home/yiyuanc/datacommons/tests/datacommons-integration-tests/custom_feature_flags.yaml) inside the `website` container at `/workspace/deploy/featureflags/custom.yaml`.
+Local integration tests run with the Spanner multi-entity schema enabled by default in the Mixer backend. This is controlled by mounting [custom_feature_flags.yaml](./custom_feature_flags.yaml) inside the `website` container at `/workspace/deploy/featureflags/custom.yaml`.
 
 The container entrypoint detects the environment variable `RESOLVE_WITH_SPANNER_EMBEDDINGS=true` (defined in `docker-compose.test.yml`) and starts Mixer using these flags.
 
 You can modify the flags inside `custom_feature_flags.yaml` if you need to tweak the Mixer server configuration for local debugging.
+
+---
+
+## GCP Sandbox Integration Tests
+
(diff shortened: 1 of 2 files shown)


clincoln8:
When you say "This test is expensive to run" do you mean like it takes a long time or that it is actually expensive in GCP?

tests/datacommons-integration-tests/README.md
clincoln8:
```suggestion
Local integration tests run with the Spanner multi-entity schema enabled by default in the Mixer backend. This is controlled by mounting [custom_feature_flags.yaml](./custom_feature_flags.yaml) inside the `website` container at `/workspace/deploy/featureflags/custom.yaml`.
```

yiyche:
> When you say "This test is expensive to run" do you mean like it takes a long time or that it is actually expensive in GCP?

@clincoln8 it takes about ~20 minutes to run because it includes terraform setup, ingestion, and embedding computation, testing NL search and teardown. 

from compute cost perspective: successful runs will cost very little but there's some risk if a run fails and the resources aren't cleaned up properly. (may need to set up some cleanup jobs to prevent this from happening)

tests/datacommons-integration-tests/run_gcp_integration_test.py
```diff
@@ -0,0 +1,905 @@
...
+    rows_count = int(db_results.get("rows", [["0"]])[0][0])
+    log_success(f"Query returned {rows_count} observations in Spanner database")
+
+    if rows_count < min_expected_rows:
+        raise RuntimeError(
+            f"Expected at least {min_expected_rows} observations, but found {rows_count} in Spanner."
+        )
+
+
+def run_serving_proxy_tests(service_name: str, region: str, project_id: str) -> None:
```
clincoln8, line 633: I'd recommend adding a /v2/observation check as well. Here's an example:

`/core/api/v2/observation?select=variable&select=entity&select=date&select=value&variable.dcids=number_of_frogs&entity.dcids=country/USA&entity.dcids=country/CAN
`

clincoln8 (approved):
Thanks!!

tests/datacommons-integration-tests/run_gcp_integration_test.py
yiyche: done. thank you!

tests/datacommons-integration-tests/README.md
yiyche: done!

PR #134, opened 2026-06-19
support batch/multi-import execution in ingestion Cloud Workflow

dwnoble:
Submit with this PR: https://github.com/datacommonsorg/import/pull/563

* Decoded Batch Imports: Added JSON decoding of input.importList in the Cloud Workflow to process batch import executions.
* Bulk Payload Construction: Added a loop (build_lists) to construct staging paths (imports_status_list), version history (imports_history_list), and update payload lists (imports_version_list) dynamically.
* Refactored Ingestion Helper Integration: Updated HTTP steps for status updating (/imports/status), version promotion (/imports/version), and history logging (/imports/ingestion-status) to submit the batch payloads.
* Dynamic Dataflow Job Naming: Suffixes multi-import Dataflow job executions with -multi to distinguish batch runs from single imports in console logs.

### DIFF
--- a/infra/dcp/modules/ingestion/workflow/main.tf
+++ b/infra/dcp/modules/ingestion/workflow/main.tf
@@ -28,18 +28,24 @@ resource "google_workflows_workflow" "ingestion_orchestrator" {
             - workflow_id: '$${sys.get_env("GOOGLE_CLOUD_WORKFLOW_EXECUTION_ID")}'
             - version: '$${"version-" + string(int(sys.now()))}'
             - bucket_name: '$${text.split(input.tempLocation, "/")[2]}'
-            - latest_version_gcs_path: '$${"gs://" + bucket_name + "/${var.ingestion_artifacts_path}/" + input.importName + "/" + version}'
             - execution_error: null
             - lock_timeout: ${var.lock_acquisition_timeout}
             - run_embeddings: ${var.enable_embeddings_generation}
             - run_postproc: ${var.enable_bigquery_postprocessing}
             - postprocessing_result: null
             - embedding_result: null
-            - sanitized_import: '$${text.replace_all(text.replace_all(text.to_lower(input.importName), "/", "-"), "_", "-")}'
+            - decoded_imports: '$${json.decode(input.importList)}'
+            - num_imports: '$${len(decoded_imports)}'
+            - first_import_name: '$${decoded_imports[0].importName}'
+            - sanitized_import: '$${text.replace_all(text.replace_all(text.to_lower(first_import_name), "/", "-"), "_", "-")}'
             - import_name_len: '$${len(sanitized_import)}'
-            - substring_end: '$${if(import_name_len < 35, import_name_len, 35)}'
+            - substring_end: '$${if(import_name_len < 30, import_name_len, 30)}'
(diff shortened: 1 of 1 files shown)


infra/dcp/modules/ingestion/workflow/main.tf
```diff
@@ -28,18 +28,24 @@ resource "google_workflows_workflow" "ingestion_orchestrator" {
...
             - lock_timeout: ${var.lock_acquisition_timeout}
             - run_embeddings: ${var.enable_embeddings_generation}
             - run_postproc: ${var.enable_bigquery_postprocessing}
             - postprocessing_result: null
             - embedding_result: null
-            - sanitized_import: '$${text.replace_all(text.replace_all(text.to_lower(input.importName), "/", "-"), "_", "-")}'
+            - decoded_imports: '$${json.decode(input.importList)}'
+            - num_imports: '$${len(decoded_imports)}'
+            - first_import_name: '$${decoded_imports[0].importName}'
+            - sanitized_import: '$${text.replace_all(text.replace_all(text.to_lower(first_import_name), "/", "-"), "_", "-")}'
```
gmechali: So this was using the importName because at the time we were only just processing that one properly.

I think the right approach here is to sanitize every import in decoded_imports. and then join them. 

The sanitized import name goes to dataflow, for which the jobs have annoying name requirements. Could you modify this piece so that if importNames = ds1/ilo, ds2/who, UNICC we end up with:
ds1_ilo_ds2_who_unicc

(note we dont care about uniqueness here, it's just a way to make the dataflow job descriptive so they know what's running! )

infra/dcp/modules/ingestion/workflow/main.tf
```diff
@@ -28,18 +28,24 @@ resource "google_workflows_workflow" "ingestion_orchestrator" {
...
             - import_name_len: '$${len(sanitized_import)}'
-            - substring_end: '$${if(import_name_len < 35, import_name_len, 35)}'
+            - substring_end: '$${if(import_name_len < 30, import_name_len, 30)}'
             - sanitized_short_import: '$${text.substring(sanitized_import, 0, substring_end)}'
-            - dataflow_job_name: '$${"${local.clean_namespace_prefix}" + sanitized_short_import + "-" + string(int(sys.now()))}'
+            - suffix: '$${if(num_imports > 1, "-multi", "")}'
```
gmechali: Could you confifm you ran this on sample imports with multi-provenances? 

This workflow was already getting really complicated. I think we need to take a TODO to clean this up, but it doesnt have to be for current milestone :)

gmechali (approved):
Thank you for handling this!

infra/dcp/modules/ingestion/workflow/main.tf
dwnoble: Updated to give the import a legible name using dashes (at jetski's recommendation)

infra/dcp/modules/ingestion/workflow/main.tf
dwnoble: I was able to successfully run the workflow with a frog dataset containing two imports/provenances, and a UN data load with two provenances.

PR #136, opened 2026-06-22
Expose Dataflow template GCS path as Terraform variable

clincoln8:
Exposes the Dataflow Flex Template container spec path as a Terraform variable `ingestion_dataflow_template_gcs_path` at the root level, freezing this as current stable `ingestion-dacf16d.json`. 

This allows overriding the template version at deployment time and we'll bump this to 1.0.1 once we have e2e working.

### DIFF
--- a/infra/dcp/modules/ingestion/helper_service/variables.tf
+++ b/infra/dcp/modules/ingestion/helper_service/variables.tf
@@ -32,8 +32,7 @@ variable "ingestion_bucket_name" {
 }
 
 variable "image" {
-  type    = string
-  default = "gcr.io/datcom-ci/datacommons-ingestion-helper:1.0.0"
+  type = string
 }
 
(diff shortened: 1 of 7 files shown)


infra/dcp/modules/ingestion/helper_service/variables.tf
```diff
@@ -32,8 +32,7 @@ variable "ingestion_bucket_name" {
 }
 
 variable "image" {
-  type    = string
-  default = "gcr.io/datcom-ci/datacommons-ingestion-helper:1.0.0"
```
clincoln8, line 36: removing this because it's redudant. the default is set at the top level here https://github.com/datacommonsorg/datacommons/blob/7ee1e79eeabd12b582f49502c8ae039bc8e3ab61/infra/dcp/variables.tf#L382

dwnoble (approved):

PR #137, opened 2026-06-22
Enable multi-entity and set default Dataflow template to 1.0.1

clincoln8:
This PR:
- Adds `enableMultiEntity: 'true'` to the Dataflow launch parameters in the Cloud Workflow definition. This enables support for multi-entity schemas during ingestion.
- Updates the default Dataflow template version to `ingestion-1.0.1.json` in the workflow module variables.

### DIFF
--- a/infra/dcp/modules/ingestion/workflow/main.tf
+++ b/infra/dcp/modules/ingestion/workflow/main.tf
@@ -48,6 +48,7 @@ resource "google_workflows_workflow" "ingestion_orchestrator" {
                 tempLocation: '$${input.tempLocation}'
                 stagingLocation: '$${input.tempLocation}'
                 forceCombineNodes: 'true'
+                isBaseDc: 'false'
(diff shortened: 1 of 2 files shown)


infra/dcp/modules/ingestion/workflow/main.tf
```diff
@@ -48,7 +48,7 @@ resource "google_workflows_workflow" "ingestion_orchestrator" {
                 tempLocation: '$${input.tempLocation}'
                 stagingLocation: '$${input.tempLocation}'
                 forceCombineNodes: 'true'
-                isBaseDc: 'false'
+                enableMultiEntity: 'true'
```
clincoln8, line 52: temp change. we will need both 

```
isBaseDc: 'false'
enableMultiEntity: 'true'
```
in stable state

clincoln8:
no longer needed

PR #145, opened 2026-06-25
allow emulator setup for partners testing DCP

yiyche:
- Configures the spanner container service to run with disable_query_null_filtered_index_check flag. This bypasses the emulator's strict optimizer checks on null-filtered indexes (like NodeEmbeddingIndex), allowing query flows to compile locally.
- Add setup script for emulator (setup_emulators.sh) and readme update
- Use MiniLM for local test


tests/datacommons-integration-tests/setup_emulators.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
dwnoble: The AI suggests we can potentially get rid of setup_emulators.sh -- what do you think?

```
Yes, you can absolutely replace or simplify the setup_emulators.sh script using Docker Compose native features.

Here are the two best ways to achieve this, along with their pros and cons.

Option 1: Native GCS Seeding via Volume Mounting (Recommended)
The GCS emulator used in the stack (fsouza/fake-gcs-server) has a built-in feature: it automatically creates buckets and seeds objects from any files/folders mounted to /data on startup.

How to implement:
Create a local directory structure for the seed data inside tests/datacommons-integration-tests/:
text
gcs_seed/
└── test-bucket/
    └── output/
        └── datacommons/
            └── nl/
                └── embeddings/
                    └── custom_catalog.yaml
Place the dummy custom_catalog.yaml file in that folder:
yaml
version: '1'
models: {}
indexes: {}
Update docker-compose.test.yml to mount this folder and set the -data flag:
yaml
gcs:
    image: fsouza/fake-gcs-server:1.47.8
    container_name: itest-gcs
    ports:
      - "9099:9099"
    command: -scheme http -port 9099 -data /data
    volumes:
      - ./gcs_seed:/data
    networks:
      - test-network
Pros:
Eliminates the setup_emulators.sh script entirely.
No curl commands, retries, or sleep loops.
Declarative and standard Docker Compose pattern.
Cons:
Requires committing the nested folder structure gcs_seed/test-bucket/... to git.
```

PR #146, opened 2026-06-26
Remove source definition from MCF and use new group syntax for pattern in sample

kmoscoe:
I've uploaded and rerun the workflow and it all works!


samples/OECD_wage_data/schema.mcf
```diff
@@ -11,7 +6,7 @@ specializationOf: dcid:dc/g/Root
 Node: dcid:OECDWages
 typeOf: dcid:Provenance
 name: "OECD Average Annual Wages"
-source: dcid:OECDSource
+source: dcid:dc/s/OrganisationForEconomicCo-operationAndDevelopmentOecd
```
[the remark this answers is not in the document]
kmoscoe: Great catch! Fixed.

samples/OECD_wage_data/config.json
```diff
@@ -17,9 +12,9 @@
       }
     },
     {
-      "pattern": "average_annual_wage.csv",
+      "pattern": "{average_annual_wage.csv,schema.mcf}",
```
gmechali, line 15: Just FYI in this config.json, you will have duplicate edges for each of the nodes in the schema.mcf. One edge associated with OECDWages nad one with OECDGenderWageGap

Is this intentional?

For cleanliness (and to serve as an example for customers) you may want to have a 
- shared_schema.mcf --> contains anything shared, will have dupe edges
- oecdwages.mcf --> contains the oecd wages provenance only + nodes only for that import
- oecdgenderwagegap.mcf --> same but for the other import.

gmechali (approved):
Approving but I do think it's worth changing the config.json a bit more :)

samples/OECD_wage_data/config.json
kmoscoe: I'm not sure exactly which edges you mean. You mean that e.g. each statvar will have edges to both the provenances? That wouldn't make sense. In the previous version I had a separate input file entry for the single schema; would the duplicate edges have resulted in that case as well? I agree that I should present the best practice, so I just want to make sure I understand (because the docs are wrong too, then.)

PR #155, opened 2026-07-08
Make versioning for DCP packages single source

yiyche:
per discussion with Gabe earlier today, there were many files that needed to be changed for each version bump. consolidating into a single VERSION file. also automaticlaly update the tf configs.

testing via 
```
gcloud builds submit \
  --config deploy/bump_version.yaml \
  --substitutions=_NEW_VERSION="beep.boop.bop",_GITHUB_BRANCH="feature/monorepo-version-management" \
  --project="datcom-ci" \
  .
```

created draft pr (https://github.com/datacommonsorg/datacommons/compare/chore/bump-version-beep.boop.bop?expand=1)


packages/datacommons-admin/datacommons_admin/__init__.py
```diff
@@ -1,3 +1,8 @@
 """CLI package for administering Data Commons instances in GCP."""
 
-__version__ = "1.1.0"
+import importlib.metadata
+
+try:
+    __version__ = importlib.metadata.version("datacommons-admin")
+except importlib.metadata.PackageNotFoundError:
```
gmechali: Is wrapping this in a try/except with a default really worth it here? 

I wonder if it's overkill?

packages/datacommons-admin/pyproject.toml
```diff
@@ -22,4 +22,4 @@ build-backend = "setuptools.build_meta"
 include-package-data = true
 
 [tool.setuptools.dynamic]
-version = {attr = "datacommons_admin.__version__"}
+version = {file = "VERSION"}
```
gmechali, line 25: could we leave this fiel unchanged and in init.py set the __version__ to read from ../../VERSION?

gmechali (dismissed):
Can you explain in the description how you've tested to make sure this works? 

I also just want to clarify that I was mostly referring to the terraform image/template versions. But it's true here it'd also require an update to so many places.

Please get a review from @dwnoble  on this too!

packages/datacommons-admin/datacommons_admin/__init__.py
yiyche: removed!

packages/datacommons-admin/pyproject.toml
yiyche: I tried that earlier, but it failed with FileNotFoundError

yiyche:
> Can you explain in the description how you've tested to make sure this works?
> 
> I also just want to clarify that I was mostly referring to the terraform image/template versions. But it's true here it'd also require an update to so many places.
> 
> Please get a review from @dwnoble on this too!

added, thanks for the feedback

deploy/bump_version.yaml
```diff
@@ -29,15 +29,24 @@ steps:
         git checkout -b $$BRANCH_NAME
 
         # Update Version Files
-        echo "Updating __init__.py files and root pyproject.toml to $$NEW_VERSION"
-        python3 -c "import glob, re; [open(f, 'w').write(re.sub(r'^__version__ = [\"\'].*[\"\']', '__version__ = \"'\"$$NEW_VERSION\"'\"', content, flags=re.MULTILINE)) for f in glob.glob('packages/*/datacommons_*/__init__.py') for content in [open(f, 'r').read()]]"
-        python3 -c "import re; [open('pyproject.toml', 'w').write(re.sub(r'^version = \".*\"', 'version = \"'\"$$NEW_VERSION\"'\"', content, flags=re.MULTILINE)) for content in [open('pyproject.toml', 'r').read()]]"
+        OLD_VERSION=$$(cat VERSION)
+        echo "Updating root VERSION file to $$NEW_VERSION"
```
dwnoble, line 33: Nice addition

dwnoble (approved):
This looks like a nice simplification-thanks @yiyche !

PR #156, opened 2026-07-08
Deprecate use of frog data + move to using spanner omni (vs cloud spanner emulator) for test

yiyche:
Primarily moving frog data (now broken for DCP) to using the [sample files](https://github.com/datacommonsorg/datacommons/tree/main/samples/OECD_wage_data). 

Previously spanner emulator was used for local testing because it was lighter weight, but we ended up just mocking the spanner client. In this version of the test, we moved to using Spanner omni which properly supports columnar. 
However, spanner omni have additional credentials requirements that requires fixes in the ingestion-helper repo (I will draft those soon, and remove these monkeypatches once those are merged). 

I do plan to make the test less schema dependent, but right now ingestion helper is just a thin wrapper around the import package. If we are able to move import code into datacommons/ repo, it would help with the duplicated / brittle code.


tests/datacommons-integration-tests/run_local_integration_test.py
```diff
@@ -190,11 +194,144 @@ def seed_gcs_emulator():
         raise RuntimeError(f"Failed to seed GCS emulator: {e}") from e  # noqa: BLE001
 
 
+def seed_local_spanner() -> None:
+    """Directly seed the Spanner emulator with the pre-processed wages dataset from JSON."""
+    import json
```
gmechali: nit: can you move all the imports to the top of the files

tests/datacommons-integration-tests/run_local_integration_test.py
```diff
@@ -190,11 +194,144 @@ def seed_gcs_emulator():
...
+    from pathlib import Path
+
+    from google.cloud import spanner
+
+    os.environ["SPANNER_EMULATOR_HOST"] = f"localhost:{SPANNER_GRPC_PORT}"
+
+    print(">>> Directly seeding Spanner emulator from JSON rows file...", flush=True)
+    spanner_client = spanner.Client(project=PROJECT_ID)
```
gmechali: just for my underrstanding, the docker-compose.test.yml is only for this local integfration test right?
you dont need any of that for the gcp_integration_test

tests/datacommons-integration-tests/run_local_integration_test.py
```diff
@@ -190,11 +194,144 @@ def seed_gcs_emulator():
         raise RuntimeError(f"Failed to seed GCS emulator: {e}") from e  # noqa: BLE001
 
 
+def seed_local_spanner() -> None:
```
gmechali: Can you explain what all this seeding is in reference to? 
Im not fully following.

tests/datacommons-integration-tests/docker-compose.test.yml
```diff
@@ -51,17 +53,16 @@ services:
...
+      - SPANNER_EMULATOR_HOST=spanner:${SPANNER_INTERNAL_GRPC_PORT:-15000}
       - STORAGE_EMULATOR_HOST=http://gcs:9099
       - OUTPUT_DIR=gs://test-bucket/output
       - RESOLVE_WITH_SPANNER_EMBEDDINGS=true
-      - PROJECT_ID=test-project
-      - GCP_PROJECT_ID=test-project
-      - GCP_SPANNER_INSTANCE_ID=test-instance
+      - PROJECT_ID=default
+      - GCP_PROJECT_ID=default
+      - GCP_SPANNER_INSTANCE_ID=default
```
gmechali, line 94: Should this image also loop in the datacommons-data image to be able to run this on our input, and run the mavn command (which essentially runs dataflow locally) for ingestion. 

Im not fully understanding the flow of this local integration test, could you epxlain it a bit pls :)

tests/datacommons-integration-tests/run_local_integration_test.py
yiyche: that's correct. the gcp test uses cloud spanner directly

tests/datacommons-integration-tests/docker-compose.test.yml
yiyche: thank you for the suggestion - i just added it. there were a few hiccups including the trailing slash, which I just used a monkeypatch to fix it for now (raised https://github.com/datacommonsorg/import/pull/620 but will need to follow up on how to test to ensure it doesn't break things) 

tldr this test 
- mvn recompiles ingestion-bundled.jar 
- boots up spanner omni + other emulators (ie gcs...) 
- datacommons-ingestion-helper initializes test-db tables using schema.sql
- ingestion helper seeds base ontology variables into Spanner
- CSV/MCF source files -- (via datacommons-data container) --> JSON-LD graph files --> GCS emulator bucket
- execute ingestion-bundled.jar for ingestion
- boot website container
- verify spanner queries + NL queries

the goal of this test is to cover the main DCP setup component to the best of my understanding (jsonld in gcs -> spanner init -> ingestion helper -> dataflow local -> serving stack / search). lmk if you have additional questions

dwnoble (approved):

PR #159, opened 2026-07-08
[DCP Preprocessor] Move the Preprocessor to inside the workflow - Part 2

gmechali:
Modify the terraform for the workflow to trigger the preprocessor job, then parse the GCS handoff file to find the graph paths to launch dataflow. 

This also modifies the permissions such that the CLI service account no longer requests permissions to start a job, instead it requests permissions to start the worfklow. And similarly, the SA for preprocessor no longer gets workflow trigger rights, but the worfklow gets the job create permission.

TODO Follow up - need to modify the ingestion history to add the preprocessor stage.

This PR also currently modifies the CLI to trigger the workflow instead of the preprocessor.


packages/datacommons-admin/datacommons_admin/ingest_cli.py
```diff
@@ -51,54 +51,43 @@ def start(imports: str | None = None) -> None:
...
     )
-    result = client.start_job(imports=imports)
+    result = client.start_workflow(imports=imports)
 
-    click.secho("Successfully started ingestion job!", fg="green", bold=True)
-    res_name = result.get("name") or result.get("metadata", {}).get("name")
+    click.secho("Successfully started ingestion workflow!", fg="green", bold=True)
+    res_name = result.get("name")
 
     if res_name:
```
dwnoble, line 75: is this check still needed? we might be able to just error out if there's no workflow name

dwnoble (approved):

packages/datacommons-admin/datacommons_admin/ingest_cli.py
gmechali: This is the execution name, not the workflow name. We already error out if the workfklow name is missing. That's above  and unmodified in this PR :)

PR #160, opened 2026-07-09
feat: consolidate Terraform image tags using root VERSION file

yiyche:
cannot put `coalesce` in variable.tf file (must be literals)

### DIFF
--- a/infra/dcp/main.tf
+++ b/infra/dcp/main.tf
@@ -59,6 +59,8 @@ resource "google_project_service" "apis" {
 }
 
 locals {
+  dcp_version = chomp(file("${path.module}/../../VERSION"))
+
   global_config = {
     project_id                    = var.project_id
     region                        = var.region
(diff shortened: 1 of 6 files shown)


infra/dcp/main.tf
```diff
@@ -85,7 +87,7 @@ locals {
 
   datacommons_services_config = {
     enable                          = var.enable_datacommons_services
-    image                           = var.datacommons_services_image
+    image                           = coalesce(var.datacommons_services_image, "gcr.io/datcom-ci/datacommons-services:${local.dcp_version}")
```
gmechali: If I understand this correctly, I can now set datacommons_services_image to "gcr.io/image/123:gabetest", but also set dcp_version 1.2.0
And in that case all iamges will be 1.2.0 except datacommons-services, is that right? 

And we support this use case for ALL images + df tempalte.

infra/dcp/main.tf
```diff
@@ -59,6 +59,8 @@ resource "google_project_service" "apis" {
 }
 
 locals {
+  dcp_version = "1.1.0"
```
gmechali: I thin kwe should also make this a variable, so that users can choose old versions if they want.

dwnoble: agreed, or they can set it to "latest"

dwnoble (approved):

infra/dcp/main.tf
yiyche: thanks! just pushed a new commit and refactored this. should be able to set it to latest now.

PR #163, opened 2026-07-14
fix(terraform): add depends_on for secrets accessor in preprocessing job

clincoln8:
### Problem
1. **Deployment Race Condition**: The Cloud Run preprocessing job creation occasionally failed with a `Permission denied` error because Terraform attempted to create the job and its Secret Manager IAM permissions in parallel.
2. **Plan-Time Evaluation Failure**: When Google Maps was disabled, we wanted to skip creating its IAM accessor. However, making it conditional using `count = var.secret_id != ""` broke `terraform plan` on clean deployments because the secret IDs (outputs from the `auth` module) were unknown at plan time.
3. **Missing DC API Key**: The preprocessing job was incorrectly using `var.datacommons_services_config.enable` to determine if it should mount `DC_API_KEY`. If serving services were disabled but ingestion was enabled, the job would run without the API key and fail.

### Solution
1. **Explicit Dependency**: Added `depends_on` to the Cloud Run job to ensure the IAM accessor permissions are applied first.
2. **Refactored Secrets Passing**: Refactored the `preprocessing_job` module to accept a list of secrets with explicit names and `enabled` flags (`var.secrets`). 
3. **Plan-Time Count Resolution**: Used `for_each` (filtered by the plan-time knowable `enabled` boolean) to conditionally create IAM accessors and mount secrets, resolving the plan-time errors.
4. **Always Enable DC API Key**: Set `enabled = true` for `DC_API_KEY` in the module call to ensure the preprocessing job always has access to it.

### DIFF
--- a/infra/dcp/modules/stack/main.tf
+++ b/infra/dcp/modules/stack/main.tf
@@ -130,16 +130,18 @@ module "ingestion_preprocessing_job" {
   use_spanner                   = true
   enable_spanner_embeddings     = var.datacommons_services_config.resolve_with_spanner_embeddings
   env_vars                      = local.cloud_run_shared_env_variables
-  secrets = {
-    dc_api_key = {
+  secrets = [
+    {
+      name      = "DC_API_KEY"
       secret_id = module.auth.dc_api_key_secret_id
       enabled   = var.datacommons_services_config.enable
-    }
-    maps_api_key = {
+    },
+    {
+      name      = "MAPS_API_KEY"
       secret_id = module.auth.maps_api_key_secret_id
       enabled   = var.datacommons_services_config.enable && !var.datacommons_services_config.website_disable_google_maps_api && (var.auth_config.google_maps_api_key != null || var.auth_config.create_google_maps_key)
     }
-  }
+  ]
 
   depends_on = [module.auth]
 }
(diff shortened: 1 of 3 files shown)


infra/dcp/modules/stack/main.tf
```diff
@@ -130,9 +130,18 @@ module "ingestion_preprocessing_job" {
...
+  secrets = [
+    {
+      name      = "DC_API_KEY"
+      secret_id = module.auth.dc_api_key_secret_id
+      enabled   = true
+    },
+    {
+      name      = "MAPS_API_KEY"
+      secret_id = module.auth.maps_api_key_secret_id
+      enabled   = var.datacommons_services_config.enable && !var.datacommons_services_config.website_disable_google_maps_api && (var.auth_config.google_maps_api_key != null || var.auth_config.create_google_maps_key)
```
gmechali: The maps API Key is actually required for the preprocessor job because we do entity resolution on their places if they use just natural language for places. So we dont REQUIRE them to use place DCIDs I believe (though I haven't fully retested this).

That's why we considered the maps API key required even if you disable it from the website.

gmechali (commented):
Overall LGTM except for making maps API key optional. It's required for now for preprocessor, but we could change that moving forward. I just haven't done that work.

infra/dcp/modules/stack/main.tf
clincoln8: okay updated to be enabled=true. 
for context, we can't rely directly on the secret_id in the count methods because that's an unknown and terraform plan throws an error. Which is why we have this enabled thingy. I guess if it's always true for now, we don't really need it, but keeping in case this changes in the future.

gmechali (approved):

PR #164, opened 2026-07-14
fix(ingestion): resolve workflow TypeError in append_lists

clincoln8:
## Problem

When starting a data ingestion workflow with multiple imports (e.g., `frog_data,bilateral_trade,OECD_wage_data`), the workflow fails at the `append_lists` step in the `build_lists` loop with the following error:

```json
{
  "failure": {
    "exception": "TypeError: in lookup, expecting dictionary, got: array\nin step \"append_lists\", routine \"main\", line: 74"
  }
}
```

The error is caused by the loop variable `imp` being evaluated as an array (list) instead of a dictionary (map) when trying to access `imp.importName` at line 74 in `workflow.yaml`.

## Underlying Cause

The variable `decoded_imports` was being constructed as a nested list (e.g., `[[A], [B], [C]]` or `[A, [B], [C]]`) instead of a flat list of maps (`[A, B, C]`). This nesting occurs because of the behavior of `list.concat` when combining lists generated by `json.decode` inside the `normalize_imports` loop.

An attempt to fix this by using native CEL list/map literals (e.g., `[{"importName": imp_str, ...}]`) failed because the GCP Workflows expression parser does not support map literals (`{...}`) directly inside expressions, leading to parse errors during deployment.

## Solution

To resolve both the nesting issue and the syntax limitations of CEL expressions, we replaced the use of `list.concat` with string concatenation inside the loops:
1.  Initialize string variables for the lists (e.g., `imports_json_str: ""`).
2.  In the loop, append each item as a JSON-formatted string fragment (e.g., `",{\"importName\":\"" + imp_str + "\"}"`).
3.  After the loop, wrap the accumulated string in `[` and `]`, and perform a single `json.decode` to construct the flat list of maps.

This approach was applied to:
-   `normalize_imports` (for `decoded_imports`)
-   `build_lists` (for `imports_status_list`, `imports_history_list`, and `imports_version_list`)

This eliminates all `list.concat` calls and nested `switch` statements inside these loops, making the workflow more robust.

TAG=agy
CONV=864b2b62-6ccd-4640-86e2-61d312cc3a6e

### DIFF
--- a/infra/dcp/modules/ingestion/workflow/workflow.yaml
+++ b/infra/dcp/modules/ingestion/workflow/workflow.yaml
@@ -26,6 +26,9 @@ main:
           - imports_status_list: []
           - imports_history_list: []
           - imports_version_list: []
+          - imports_status_str: ""
+          - imports_history_str: ""
+          - imports_version_str: ""
           - launch_params:
               projectId: '$${project_id}'
               spannerInstanceId: '$${default(map.get(input, "spannerInstanceId"), "")}'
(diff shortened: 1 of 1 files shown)


gmechali:
oh nice catch, thanks for the fix!

gmechali (approved):

PR #174, opened 2026-07-17
feat(cli): support remote GCS state fetching

dwnoble:
Adds support for fetching Terraform state outputs directly from GCS, allowing DCP admin CLI commands to be executed from anywhere outside of the local initialized Terraform folder.

### Key Changes:
- Added global options under the `admin` command group:
  - `--project-id`: GCP project ID used to locate the state bucket.
  - `--instance-name`: DCP instance name used to locate the state bucket.
  - `--tf-state-location`: Direct GCS URI pointing to the Terraform state file.
- Context-Aware Remote State Fetching:
  - Updated `get_terraform_output` to check Click context parameters. If specified, it fetches the state JSON from GCS and extracts values directly. If not, it falls back to local `terraform output -json`.
- Colorized Error Monkeypatch:
  - Custom styled Click exceptions to colorize all CLI error outputs to stderr in bold red.

### Usage Examples:
Execute configuration tasks from outside your deployment directory:

```bash
# Example A: Query ingestion configuration using project ID and instance name
datacommons admin \
  --project-id datcom-website-dev \
  --instance-name dan-ns \
  ingest show-config

# Example B: Initialize database using a direct state location
datacommons admin \
  --tf-state-location gs://tf-state-dan-ns-datcom-website-dev/terraform/state/dan-ns/default.tfstate \
  init-db
```

*(Note: Chained off 'deprecate-namespace' branch/PR)*

### DIFF
--- a/packages/datacommons-admin/datacommons_admin/admin_cli.py
+++ b/packages/datacommons-admin/datacommons_admin/admin_cli.py
@@ -35,14 +35,14 @@ GITHUB_RAW_BASE_URL = "https://raw.githubusercontent.com/datacommonsorg/datacomm
 GITHUB_REPO_URL = "https://github.com/datacommonsorg/datacommons.git"
 
 
-def _get_default_bucket_name(namespace: str, project_id: str) -> str:
+def _get_default_bucket_name(instance_name: str, project_id: str) -> str:
     """Returns the default Google Cloud Storage bucket name for Terraform state."""
-    return f"tf-state-{namespace}-{project_id}"
+    return f"tf-state-{instance_name}-{project_id}"
 
 
-def _get_default_state_prefix(namespace: str) -> str:
+def _get_default_state_prefix(instance_name: str) -> str:
     """Returns the default Google Cloud Storage object prefix for Terraform state."""
-    return f"terraform/state/{namespace}"
+    return f"terraform/state/{instance_name}"
 
 
 def _log_resolved_value(label: str, value: str, is_default: bool, indent: int = 2):
@@ -229,30 +229,30 @@ def admin() -> None:
     """Manage a Data Commons Platform instance in Google Cloud"""
 
 
-def _validate_namespace(ns: str) -> Tuple[bool, str]:
-    if not ns:
-        return False, "Namespace must not be empty."
-    if len(ns) > 16:
+def _validate_instance_name(name: str) -> Tuple[bool, str]:
+    if not name:
+        return False, "Instance name must not be empty."
+    if len(name) > 16:
         return (
             False,
-            f"Namespace must be 16 characters or less (currently {len(ns)} characters).",
+            f"Instance name must be 16 characters or less (currently {len(name)} characters).",
         )
-    if not re.match(r"^[a-z]([-a-z0-9]*[a-z0-9])?$", ns):
+    if not re.match(r"^[a-z]([-a-z0-9]*[a-z0-9])?$", name):
         return (
             False,
-            "Namespace must start with a lowercase letter, end with a lowercase letter or number, and contain only lowercase alphanumeric characters and dashes.",
+            "Instance name must start with a lowercase letter, end with a lowercase letter or number, and contain only lowercase alphanumeric characters and dashes.",
         )
     return True, ""
 
 
 def _resolve_project_config(
-    project_id: str, namespace: str, force: bool
+    project_id: str, instance_name: str, force: bool
 ) -> Tuple[str, str, Path]:
-    """Resolves project ID and namespace, and determines target directory."""
+    """Resolves project ID and instance name, and determines target directory."""
     if project_id:
         _log_resolved_value("Project ID", project_id, is_default=False)
-    if namespace:
-        _log_resolved_value("Namespace", namespace, is_default=False)
+    if instance_name:
+        _log_resolved_value("Instance Name", instance_name, is_default=False)
 
     resolved_project_id = project_id.strip()
     if not resolved_project_id:
@@ -262,34 +262,34 @@ def _resolve_project_config(
     if not resolved_project_id:
         raise click.ClickException("GCP project ID must not be empty.")
 
-    resolved_namespace = namespace.strip()
-    if resolved_namespace:
-        is_valid, err_msg = _validate_namespace(resolved_namespace)
+    resolved_instance_name = instance_name.strip()
+    if resolved_instance_name:
+        is_valid, err_msg = _validate_instance_name(resolved_instance_name)
         if not is_valid:
             raise click.ClickException(err_msg)
 
     while True:
-        if not resolved_namespace:
-            resolved_namespace = _prompt("Namespace", type=str).strip()
-            is_valid, err_msg = _validate_namespace(resolved_namespace)
+        if not resolved_instance_name:
+            resolved_instance_name = _prompt("Instance Name", type=str).strip()
+            is_valid, err_msg = _validate_instance_name(resolved_instance_name)
             if not is_valid:
                 click.secho(f"Error: {err_msg}", fg="red")
-                resolved_namespace = ""
+                resolved_instance_name = ""
                 continue
 
-        target_dir = Path.cwd() / resolved_namespace
+        target_dir = Path.cwd() / resolved_instance_name
         if target_dir.exists() and not force:
             click.secho(
-                f"Error: Folder '{resolved_namespace}' already exists locally. "
-                "Please specify a different namespace, or use --force to overwrite.",
+                f"Error: Folder '{resolved_instance_name}' already exists locally. "
+                "Please specify a different instance name, or use --force to overwrite.",
                 fg="yellow",
             )
-            resolved_namespace = ""
+            resolved_instance_name = ""
             continue
 
         break
 
-    return resolved_project_id, resolved_namespace, target_dir
+    return resolved_project_id, resolved_instance_name, target_dir
 
 
 def _check_existing_files(target_dir: Path, use_remote_state: bool, force: bool):
@@ -315,7 +315,7 @@ def _check_existing_files(target_dir: Path, use_remote_state: bool, force: bool)
 def _setup_dcp_config_dir(
     target_dir: Path,
     project_id: str,
-    namespace: str,
+    instance_name: str,
     bucket_name: str,
     tf_state_prefix: str,
     dc_api_key: str,
@@ -364,7 +364,8 @@ def _setup_dcp_config_dir(
         # Modify tfvars_example with actual values
         tfvars_content = tfvars_example
         tfvars_content = tfvars_content.replace('"$$PROJECT_ID$$"', f'"{project_id}"')
-        tfvars_content = tfvars_content.replace('"$$NAMESPACE$$"', f'"{namespace}"')
+        tfvars_content = tfvars_content.replace('"$$NAMESPACE$$"', f'"{instance_name}"')
+        tfvars_content = tfvars_content.replace('"$$INSTANCE_NAME$$"', f'"{instance_name}"')
         if api_key:
             tfvars_content = tfvars_content.replace('"$$DC_API_KEY$$"', f'"{api_key}"')
 
@@ -401,7 +402,14 @@ def _setup_dcp_config_dir(
     help="Google Cloud Platform project ID used for all resources related to your Data Commons instance.",
 )
 @click.option(
-    "--namespace", default="", help="Namespace prefix for provisioned resources."
+    "--instance-name",
+    default="",
+    help="Unique identifier used as a prefix for resource naming (replaces --namespace).",
+)
+@click.option(
+    "--namespace",
+    default="",
+    help="DEPRECATED: Use --instance-name instead.",
 )
 @click.option("--dc-api-key", default="", help="Data Commons API key.")
 @click.option(
@@ -432,11 +440,12 @@ def _setup_dcp_config_dir(
 @click.option(
     "--tf-state-prefix",
     default="",
-    help="Google Cloud Storage object prefix for Terraform state file (default: terraform/state/{namespace}).",
+    help="Google Cloud Storage object prefix for Terraform state file (default: terraform/state/{instance_name}).",
 )
 def init(
     project_id: str,
     namespace: str,
+    instance_name: str,
     dc_api_key: str,
     tf_git_ref: str,
     force: bool,
@@ -448,11 +457,14 @@ def init(
     """Initialize Terraform scaffolding for Data Commons administration/infrastructure."""
     click.secho("Data Commons Admin Init", fg="cyan", bold=True)
 
+    # For backward compatibility, fallback to namespace if instance_name is empty
+    effective_instance_name = instance_name.strip() or namespace.strip()
+
     # 1. Project Configs
     click.secho("\n[Project configuration]", fg="cyan", bold=True)
     click.secho("Configuring project settings...", fg="bright_black")
-    resolved_project_id, resolved_namespace, target_dir = _resolve_project_config(
-        project_id, namespace, force
+    resolved_project_id, resolved_instance_name, target_dir = _resolve_project_config(
+        project_id, effective_instance_name, force
     )
 
     # 2. Terraform Setup
@@ -467,7 +479,7 @@ def init(
     resolved_bucket_name = (
         _configure_remote_state(
             resolved_project_id,
-            resolved_namespace,
+            resolved_instance_name,
             tf_state_bucket,
             tf_state_bucket_location,
         )
@@ -476,7 +488,7 @@ def init(
     )
 
     resolved_tf_state_prefix = tf_state_prefix.strip() or _get_default_state_prefix(
-        resolved_namespace
+        resolved_instance_name
     )
 
     # 3. DCP config dir setup
@@ -485,7 +497,7 @@ def init(
     _setup_dcp_config_dir(
         target_dir,
         resolved_project_id,
-        resolved_namespace,
+        resolved_instance_name,
         resolved_bucket_name,
         resolved_tf_state_prefix,
         dc_api_key,
@@ -494,7 +506,7 @@ def init(
     )
 
     click.secho(
-        f"Customize variables in {resolved_namespace}/terraform.tfvars as needed.",
+        f"Customize variables in {resolved_instance_name}/terraform.tfvars as needed.",
         fg="green",
     )
     click.secho(
--- a/packages/datacommons-admin/datacommons_admin/admin_cli.py
+++ b/packages/datacommons-admin/datacommons_admin/admin_cli.py
@@ -225,8 +225,33 @@ def _get_github_templates(ref: str) -> tuple[str, str, str, str]:
 
 
 @click.group()
-def admin() -> None:
+@click.option(
+    "--project-id",
+    default=None,
+    help="GCP project ID used to locate the remote Terraform state bucket.",
+)
+@click.option(
+    "--instance-name",
+    default=None,
+    help="DCP instance name (prefix) used to locate the remote Terraform state bucket.",
+)
+@click.option(
+    "--tf-state-location",
+    default=None,
+    help="Direct GCS URI pointing to the Terraform state file (gs://bucket/prefix/default.tfstate).",
+)
+@click.pass_context
+def admin(
+    ctx: click.Context,
+    project_id: str | None,
+    instance_name: str | None,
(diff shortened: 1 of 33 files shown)


packages/datacommons-admin/datacommons_admin/admin_cli.py
```diff
@@ -364,7 +389,8 @@ def _setup_dcp_config_dir(
         # Modify tfvars_example with actual values
         tfvars_content = tfvars_example
         tfvars_content = tfvars_content.replace('"$$PROJECT_ID$$"', f'"{project_id}"')
-        tfvars_content = tfvars_content.replace('"$$NAMESPACE$$"', f'"{namespace}"')
+        tfvars_content = tfvars_content.replace('"$$NAMESPACE$$"', f'"{instance_name}"')
```
gmechali: Any reason to leave the NAMESPACE here?

packages/datacommons-admin/datacommons_admin/admin_cli.py
```diff
@@ -448,11 +482,14 @@ def init(
     """Initialize Terraform scaffolding for Data Commons administration/infrastructure."""
     click.secho("Data Commons Admin Init", fg="cyan", bold=True)
 
+    # For backward compatibility, fallback to namespace if instance_name is empty
```
gmechali: I think we rip the bandaid off tbh. Or we ll carry this over far longer than we need.
No one is really using this yet, if so, they can continue using previous version of the CLI.

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
```diff
@@ -31,60 +31,154 @@
...
-    if not shutil.which("terraform"):
-        raise click.ClickException(
-            "Terraform CLI not found. Please ensure Terraform is installed and available in your PATH."
-        )
+def _get_outputs_from_gcs(
+    tf_state_location: str | None,
+    project_id: str | None,
+    instance_name: str | None,
+) -> dict:
+    from google.cloud import storage
```
gmechali: nit: move imports to top of file?

packages/datacommons-admin/datacommons_admin/tf_utils.py
```diff
@@ -31,60 +31,154 @@
...
+                project_id = project_id or cur_ctx.obj.get("project_id")
+                instance_name = instance_name or cur_ctx.obj.get("instance_name")
+                tf_state_location = tf_state_location or cur_ctx.obj.get("tf_state_location")
+            cur_ctx = cur_ctx.parent
+
+    use_remote = bool(tf_state_location or (project_id and instance_name))
+
+    if use_remote:
+        outputs = _get_outputs_from_gcs(tf_state_location, project_id, instance_name)
+    else:
```
gmechali: can we move this elsewhere and put it in a helper for validation to use the CLI (or at least this function)? 

There's no point in fetching the terraform state and doig all that logic above if it's gonna fail here.

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
```diff
@@ -31,60 +31,154 @@
...
-        outputs = json.loads(result.stdout)
-    except json.JSONDecodeError:
-        raise click.ClickException(
-            "Failed to parse 'terraform output -json'. The output was not valid JSON."
-        )
+        bucket = client.bucket(bucket_name)
+        blob = bucket.blob(blob_name)
+        
+        try:
+            content = blob.download_as_text()
```
gmechali: one criticism from Jorge is that he doesn't want us to bring in the terraform state locally as it opens up the possibility of someone running a tf command they shouldnt.

For this specific scenario, we can just read the file from GCS and parse the output without copying the data over in the local terraform state right?

packages/datacommons-admin/datacommons_admin/tf_utils.py
dwnoble: Moved context parsing into `_resolve_remote_state_params()` and local execution checks into `_get_outputs_from_local()`.

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
dwnoble: Yes, this reads the GCS state JSON directly into memory without writing `.tfstate` to disk or running local terraform commands.

dwnoble:
Addressed review feedback: added whitespace stripping, passed `project` to `storage.Client`, added dict type checks for state JSON, and added unit tests in `test_tf_utils.py`.

packages/datacommons-admin/datacommons_admin/admin_cli.py
```diff
@@ -20,8 +20,33 @@
 
 
 @click.group()
-def admin() -> None:
+@click.option(
```
gmechali, line 23: Do you think this would be better if it was another way to trigger the init command? Like

datacommons admin init-remote --project --tfstate.... 

And it could just be another command in the init_cli, because essentially this is a way to initialize your local environment right? 
https://github.com/datacommonsorg/datacommons/blob/main/packages/datacommons-admin/datacommons_admin/init/init_cli.py

packages/datacommons-admin/datacommons_admin/admin_cli.py
```diff
@@ -20,8 +20,33 @@
 
 
 @click.group()
```
gmechali, line 22: Im not sure im fully understanding how you're using this is. 

Is it whenever I run an admin command, if we provide these flags, it ll try to connect to the remote state?

How should we handle the lack of terraform.tfvars?
For our testbed example, we push that as a secret, but that isnt  required for DCP instances, so we could be bringing in the remote state without a tfvars file. Is that an issue?

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
```diff
@@ -31,8 +32,120 @@
 TF_OUTPUT_INGESTION_WORKFLOW_NAME = "ingestion_workflow_name"
 
 
-def get_terraform_output(key: str) -> str:
-    """Fetches a specific key from `terraform output -json` with graceful error handling."""
+def _resolve_remote_state_params() -> tuple[str | None, str | None, str | None]:
```
gmechali: Wdyt about using pydantic models to represent the input params on each package, like adminParams, then passing that around, so you can just read adminParams.projectId?

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
```diff
@@ -31,8 +32,120 @@
...
+
+    p_id = project_id.strip() if project_id and project_id.strip() else None
+    i_name = instance_name.strip() if instance_name and instance_name.strip() else None
+    state_loc = (
+        tf_state_location.strip()
+        if tf_state_location and tf_state_location.strip()
+        else None
+    )
+
+    return p_id, i_name, state_loc
```
gmechali: similarly to comment above, this response type tuple[str | None, str | None, str | None] makes it pretty hard to understand what s going on right off the bat, and obviously int he future could grow easily by 2/3 more.

Could this be another model

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
```diff
@@ -31,8 +32,120 @@
...
+
+    # Resolve the GCS URI
+    if tf_state_location:
+        gcs_uri = tf_state_location
+        if not gcs_uri.endswith(".tfstate"):
+            gcs_uri = gcs_uri.rstrip("/") + "/default.tfstate"
+    elif project_id and instance_name:
+        gcs_uri = f"gs://tf-state-{instance_name}-{project_id}/terraform/state/{instance_name}/default.tfstate"
+    else:
+        return {}
```
gmechali: error instead of return empty?

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
```diff
@@ -31,8 +32,120 @@
...
+        parts = gcs_uri[5:].split("/", 1)
+        if len(parts) < 2 or not parts[1]:
+            raise click.ClickException(
+                f"Invalid GCS URI '{gcs_uri}'. Must specify bucket and object path."
+            )
+        bucket_name, blob_name = parts[0], parts[1]
+
+        try:
+            client = (
+                storage.Client(project=project_id) if project_id else storage.Client()
```
gmechali: wouldnt we make project ID required for this?

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
```diff
@@ -31,8 +32,120 @@
...
+                f"Invalid GCS URI '{gcs_uri}'. Must start with 'gs://'."
+            )
+        parts = gcs_uri[5:].split("/", 1)
+        if len(parts) < 2 or not parts[1]:
+            raise click.ClickException(
+                f"Invalid GCS URI '{gcs_uri}'. Must specify bucket and object path."
+            )
+        bucket_name, blob_name = parts[0], parts[1]
+
+        try:
```
gmechali: Why the doubly nested try statements?

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
dwnoble: When `--tf-state-location` is provided, `project_id` is optional because the URI is fully specified. When constructing the default URI, both `--project-id` and `--instance-name` are required.

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
dwnoble: Flattened the exception handling into single linear try/except blocks.

packages/datacommons-admin/datacommons_admin/admin_cli.py
dwnoble: We don't actually need the secrets in this case. We're using the remote terraform state to get the ingestion helper URL & other resource URLs.

Pulling the state from the terraform GCS bucket is an added layer of safety because it pulls what is actually deployed, where a local terraform.tfvars could be out of sync with what's actually in GCP.

packages/datacommons-admin/datacommons_admin/admin_cli.py
dwnoble: can you tell me more what you are thinking the behavior of an `init-remote` command would do?

i like the idea of having a recommended or automated way to sync the tfvars file with our deployment, but i think it's also a separate issue from pulling the state of what's currently deployed in a DCP instance based on the terraform output stored in GCS, regardless of where the tfvars are stored.

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
```diff
@@ -31,60 +31,154 @@
...
+    instance_name: str | None,
+) -> dict:
+    from google.cloud import storage
+    from google.cloud.exceptions import GoogleCloudError
+
+    # Resolve the GCS URI
+    if tf_state_location:
+        gcs_uri = tf_state_location.strip()
+        if not gcs_uri.endswith(".tfstate"):
+            gcs_uri = gcs_uri.rstrip("/") + "/default.tfstate"
```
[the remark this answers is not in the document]
dwnoble: Addressed by normalizing CLI strings before constructing `TerraformStateConfig`.

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
```diff
@@ -31,60 +31,154 @@
...
+        if not gcs_uri.startswith("gs://"):
+            raise click.ClickException(f"Invalid GCS URI '{gcs_uri}'. Must start with 'gs://'.")
+        parts = gcs_uri[5:].split("/", 1)
+        if len(parts) < 2 or not parts[1]:
+            raise click.ClickException(f"Invalid GCS URI '{gcs_uri}'. Must specify bucket and object path.")
+        bucket_name, blob_name = parts[0], parts[1]
+
+        try:
+            client = storage.Client()
+        except Exception as e:
```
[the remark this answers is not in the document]
dwnoble: Addressed. The storage client receives `project_id` when it is available.

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
```diff
@@ -31,60 +31,154 @@
...
+            )
 
-    if not outputs:
-        from pathlib import Path
+        try:
+            state_data = json.loads(content)
+        except json.JSONDecodeError:
+            raise click.ClickException(f"Failed to parse Terraform state at '{gcs_uri}' as valid JSON.")
+
+        outputs = state_data.get("outputs", {})
```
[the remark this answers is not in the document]
dwnoble: Addressed. Parsed state is validated as a JSON object before reading `outputs`.

packages/datacommons-admin/datacommons_admin/tf_utils.py
```diff
@@ -31,60 +31,154 @@
...
+            else:
+                raise click.ClickException(
+                    f"No Terraform outputs found in '{cwd}'.\n"
+                    "Please ensure you have successfully run 'terraform apply' to generate the deployment state."
+                )
+
     if key not in outputs:
         from pathlib import Path
-
+        location_desc = tf_state_location or f"GCP project: '{project_id}' / instance: '{instance_name}'" if use_remote else f"'{Path.cwd()}'"
```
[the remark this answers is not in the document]
dwnoble: Addressed by moving this logic into the explicit `location_description` property.

packages/datacommons-admin/tests/test_admin_cli.py
```diff
@@ -540,3 +580,64 @@ def test_ingest_start_with_imports_success(
...
+        [
+            "--project-id", "mock-project",
+            "--instance-name", "mock-instance",
+            "test-get-output"
+        ]
+    )
+    assert result.exit_code == 0
+    assert "VAL=gcs-resolved-val" in result.output
+    mock_client_inst.bucket.assert_called_once_with("tf-state-mock-instance-mock-project")
+    mock_bucket.blob.assert_called_once_with("terraform/state/mock-instance/default.tfstate")
```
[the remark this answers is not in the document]
dwnoble: Added coverage verifying the expected project ID is passed to `storage.Client`.

packages/datacommons-admin/tests/test_admin_cli.py
```diff
@@ -540,3 +580,64 @@ def test_ingest_start_with_imports_success(
...
+
+    # Test with tf-state-location
+    mock_client_inst.reset_mock()
+    mock_bucket.reset_mock()
+    result = runner.invoke(
```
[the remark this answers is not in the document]
dwnoble: Addressed—the GCS mocks are reset before testing explicit state-location behavior.

packages/datacommons-admin/tests/test_admin_cli.py
```diff
@@ -540,3 +580,64 @@ def test_ingest_start_with_imports_success(
...
+        admin,
+        [
+            "--tf-state-location", "gs://custom-bucket/custom-prefix/state.tfstate",
+            "test-get-output"
+        ]
+    )
+    assert result.exit_code == 0
+    assert "VAL=gcs-resolved-val" in result.output
+    mock_client_inst.bucket.assert_called_once_with("custom-bucket")
+    mock_bucket.blob.assert_called_once_with("custom-prefix/state.tfstate")
```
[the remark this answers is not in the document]
dwnoble: Added in `a1e0d86`. The test now verifies that URI-only resolution initializes `storage.Client` without a project override.

packages/datacommons-admin/datacommons_admin/admin_cli.py
dwnoble: Agreed—the obsolete namespace compatibility path is no longer part of the current change.

packages/datacommons-admin/datacommons_admin/admin_cli.py
dwnoble: Agreed. The current implementation no longer carries the previous namespace compatibility path.

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
dwnoble: Done—the GCS imports now live at module scope.

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
dwnoble: Good call on using a named configuration object. I used a small frozen dataclass rather than Pydantic because these are three already-parsed Click values and do not require schema coercion.

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
dwnoble: Addressed with `TerraformStateConfig`, giving the parameters named fields without depending on tuple ordering.

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
dwnoble: Addressed. Invalid partial remote configuration now raises a clear `ClickException`; no remote options explicitly selects local mode.

packages/datacommons-admin/datacommons_admin/admin_cli.py
dwnoble: Thanks for raising this. After tightening the behavior, I think this fits better as admin-level state-source selection than an init command. It does not initialize or synchronize a local environment; it only selects the deployed state used by the requested admin operation. The CLI help now makes that distinction explicit.

packages/datacommons-admin/datacommons_admin/admin_cli.py
dwnoble: Yes, the remote options apply to the requested admin command whenever it needs Terraform outputs. A local `terraform.tfvars` is not required because the command reads deployed outputs directly from GCS and never creates or updates local Terraform files. I clarified this behavior in the admin help text.

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
```diff
@@ -30,37 +36,206 @@
...
+def _clean_str(value: object | None) -> str | None:
+    """Strips whitespace from string values and normalizes empty strings to None."""
+    if isinstance(value, str):
+        cleaned = value.strip()
+        return cleaned if cleaned else None
+    return None
+
+
+@dataclass(frozen=True)
+class TerraformStateConfig:
```
gmechali: WDYT about moving this to a model/ subdir?

packages/datacommons-admin/tests/core/test_tf_utils.py
```diff
@@ -0,0 +1,360 @@
...
+    assert "VAL=gcs-resolved-val" in result.output
+    mock_storage_client.assert_called_with(project="mock-project")
+    mock_client_inst.bucket.assert_called_once_with(
+        "tf-state-mock-instance-mock-project"
+    )
+    mock_bucket.blob.assert_called_once_with(
+        "terraform/state/mock-instance/default.tfstate"
+    )
+
+    # Test with tf-state-location
```
gmechali: should the lines below this just be a separate test where you reset the mocks between each test?
Better practice not two test for two different things wihtin one test.

gmechali (approved):
Few smaller comments, but otherwise LGTM, thanks Dan! 

Can you also maybe add something in the readme about how to use this? (or maybe we should just do a DC CLI Cheatsheet somewhere?)

packages/datacommons-admin/datacommons_admin/core/utils/tf_utils.py
dwnoble: done

packages/datacommons-admin/tests/core/test_tf_utils.py
dwnoble: done

dwnoble:
> Few smaller comments, but otherwise LGTM, thanks Dan!
> 
> Can you also maybe add something in the readme about how to use this? (or maybe we should just do a DC CLI Cheatsheet somewhere?)

Added to the packages/datacommons-cli/README.md doc and linked that from the main readme.md

Submitting to get this one in, but happy to revisit the docs if you have feedback on them

PR #186, opened 2026-07-24
Expose Data Commons serving service account email

tillywoodfield:
Expose the email of the Data Commons serving SA at the module root so that consumers can reference it when provisioning other infrastructure.

### DIFF
--- a/infra/dcp/modules/stack/outputs.tf
+++ b/infra/dcp/modules/stack/outputs.tf
@@ -17,7 +17,8 @@ output "datacommons_service_name" {
 }
 
 output "datacommons_service_service_account_email" {
-  value = length(module.datacommons_services) > 0 ? module.datacommons_services[0].service_account_email : null
+  description = "Email of the service account used by the Data Commons serving service"
+  value       = length(module.datacommons_services) > 0 ? module.datacommons_services[0].service_account_email : null
 }
 
 output "ingestion_workflow_id" {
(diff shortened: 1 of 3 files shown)


infra/dcp/modules/stack/outputs.tf
```diff
@@ -16,6 +16,10 @@ output "datacommons_service_name" {
   value = length(module.datacommons_services) > 0 ? module.datacommons_services[0].service_name : null
 }
 
+output "datacommons_service_service_account_email" {
+  value = length(module.datacommons_services) > 0 ? module.datacommons_services[0].service_account_email : null
+}
```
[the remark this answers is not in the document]
tillywoodfield: Done

infra/dcp/modules/stack/outputs.tf
```diff
@@ -16,6 +16,11 @@ output "datacommons_service_name" {
   value = length(module.datacommons_services) > 0 ? module.datacommons_services[0].service_name : null
 }
 
+output "datacommons_service_service_account_email" {
+  description = "Email of the service account used by the serving service"
```
juliawu: For consistency, how about we also use the same language in this description as the others?
```suggestion
  description = "Email of the service account used by the Data Commons serving service"
```

juliawu (approved):
Hello and welcome to the Repo! Adding the service account email to the output is a great idea, and this generally looks good to me. I've just got one minor suggestion.

infra/dcp/modules/stack/outputs.tf
tillywoodfield: @juliawu thanks! Have updated

PR #191, opened 2026-07-24
[Retention PEriod] Creates more buffer for safety. 

gmechali:
We could bring it down later if rollback is enabled and works well.


clincoln8 (approved):
Sometimes we set default in modules/stack/variables.tf too. I'm not totally sure when that's also necessary...

gmechali:
yeah ive got the default in both rn. lets rediscuss later

PR #196, opened 2026-07-29
fix(tests): update GCS JSON-LD blob discovery for provenance paths

clincoln8:
### Summary
Fixes the hermetic integration test runner and aligns the Spanner node golden file so integration tests pass reliably and deterministically in offline and pull request environments without requiring external API secrets.

---

### Problems & Root Causes

1. **GCS Path Discovery in Test Runner:**
   * `run_spanner_loader()` in `run_local_integration_test.py` used a rigid path filter (`if 'output/jsonld/' in name` with a 5-part path split requirement).
   * After the preprocessor export directory structure was updated to output per-provenance directories (`output/jsonld/<import_name>_<timestamp>/*/*.jsonld`), the test harness failed to discover generated `.jsonld` files, preventing them from being passed to the Spanner loader.

2. **External Dependency in Hermetic Node Golden:**
   * `golden/spanner_nodes.json` included 38 external country nodes (`country/AUS`, `country/USA`, etc.).
   * During ingestion, `datacommons-data-processor` (in `dcpbridge` mode) attempts to resolve country entities by making an outbound network call to the online Data Commons Recon API (`https://api.datacommons.org/v2/resolve`), which requires a valid `DC_API_KEY`.
   * In GitHub Actions PR runs (specifically PRs originating from forks), GitHub's security model deliberately withholds repository secrets, resulting in an empty/missing `DC_API_KEY`.
   * When entity resolution fails due to the missing key (HTTP 401), the preprocessor gracefully skips generating `Node` definitions for the unresolved external country entities. While observations and serving tests continue to function, `test_spanner_node_seeding` asserted exact equality against a golden file containing these 38 country nodes, causing a test failure.

---

### Fix & Changes

1. **Dynamic GCS Directory Discovery:**
   * Updated `run_spanner_loader()` to dynamically inspect the GCS emulator bucket for `.jsonld` files, resolve their top-level import directory (e.g. `gs://test-bucket/output/jsonld/wages_<timestamp>/*/*.jsonld`), and pass them to the Spanner loader container.

2. **Hermetic Golden Alignment:**
   * Removed the 38 external `country/*` node entries from `golden/spanner_nodes.json`.
   * The golden now strictly asserts the 21 deterministic, self-contained schema nodes:
     - Base ontology classes seeded by `seed_database` (`Class`, `StatisticalVariable`, `StatVarObservation`, `StatVarGroup`, `Topic`, `dc/g/Root`).
     - Custom dataset schema and provenance nodes from the sample MCF files (`average_annual_wage`, `gender_wage_gap`, `OECDGroup`, `OECDWages`, `OECDGenderWageGap`, and property triples).
   * This ensures the integration test suite is truly hermetic and does not fail on PRs due to absent network credentials.


gmechali (commented):
thanks! And sorry for breaking :(

gmechali (approved):

gmechali:
why are we running these in github actions rather than Cloud Build like everything else in our CI?

Removing the proxy nodes sgtm for now, especially since we hope to remove that feature entirely within a few weeks.

gmechali (commented):
reapproving, thanks christie!

gmechali (approved):

clincoln8:
> why are we running these in github actions rather than Cloud Build like everything else in our CI?

That's a good question, let's revisit this as we continue to iterate on the long term cicd plans!

PR #199, opened 2026-07-30
[DCP Ingestion] Fix lock acquisition retry following a change in error code.

gmechali:
Error code was changed from 400 to 503 in https://github.com/datacommonsorg/import/pull/697/changes

It  broke the retry logic, this will fix it. I wonder if there's a better error code to use though? Food for thought.


clincoln8 (approved):
Agreed 503 isn't the most helpful error here. I'll add to the list of things to revisit

gmechali:
HTTP 423 means locked, so we may have our answer 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/423

PR #208, opened 2026-08-12
remove trailing spaces from docs

pranj-s:
testing jetski --> github dev workflow.

note: breaking integ test is non-blocking and unrelated to this doc change.


gmechali (approved):
Thanks for the cleanup!

PR #209, opened 2026-08-13
feat(infra): add cost optimization and performance variables for Cloud Run and Redis VPC connector

saurabh-google:
## Motivation
Custom Data Commons setups already provide knobs like `cpu_idle` to optimize infrastructure costs for dev and low-traffic environments. Bringing these same customization options to DCP enables teams to reduce baseline cloud spend while maintaining fine-grained control over performance.

## Summary
This PR exposes configurable Terraform variables to enable cost optimization and performance tuning for DCP (Data Commons Platform) deployments.

## Changes
1. **Cloud Run Resource Controls (`cpu_idle` & `startup_cpu_boost`)**:
   - Added `cpu_idle` (default `false`) and `startup_cpu_boost` (default `true`) configurations to `datacommons_services` and `ingestion/helper_service` Cloud Run services.
   - Setting `cpu_idle = true` allows CPU to only be allocated during request processing, reducing running costs for lower-traffic environments.
   - Enabling `startup_cpu_boost` mitigates cold start latencies during container boots.

2. **Redis VPC Access Connector Scaling**:
   - Parameterized `vpc_connector_min_instances` (default `2`) and `vpc_connector_max_instances` (default `10`) in the `redis` module.
   - Allows deployments to adjust baseline VPC connector instance counts to optimize infrastructure costs.

3. **Templates & Variables**:
   - Added commented examples in `terraform.tfvars.template`.
   - Threaded configurations through root `variables.tf`, `main.tf`, and the `stack` module with backwards-compatible defaults.

## Testing
- Validated Terraform syntax and module wiring across modified modules.

### DIFF
--- a/infra/dcp/modules/stack/variables.tf
+++ b/infra/dcp/modules/stack/variables.tf
@@ -80,8 +80,8 @@ variable "redis_config" {
     replica_count               = number
     vpc_network_name            = string
     vpc_connector_cidr          = string
-    vpc_connector_min_instances = optional(number, 2)
-    vpc_connector_max_instances = optional(number, 10)
+    vpc_connector_min_instances = number
+    vpc_connector_max_instances = number
   })
 }
 
--- a/infra/dcp/modules/stack/variables.tf
+++ b/infra/dcp/modules/stack/variables.tf
@@ -110,6 +110,8 @@ variable "ingestion_config" {
     # Workflow & Helper Service
     workflow_lock_acquisition_timeout = number
     helper_service_image              = optional(string)
+    helper_service_cpu_idle           = optional(bool)
+    helper_service_startup_cpu_boost  = optional(bool)
 
     # Dataflow network configuration
     # Use WORKER_IP_PRIVATE when a compute.vmExternalIpAccess org policy
--- a/infra/dcp/modules/stack/variables.tf
+++ b/infra/dcp/modules/stack/variables.tf
@@ -45,6 +45,8 @@ variable "datacommons_services_config" {
     max_instances                   = number
     cpu                             = string
     memory                          = string
+    cpu_idle                        = optional(bool, false)
+    startup_cpu_boost               = optional(bool, true)
     google_analytics_tag            = string
     enable_mcp                      = bool
     search_scope                    = string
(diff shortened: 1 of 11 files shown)


dwnoble:
Thank you @saurabh-google !

Were you able to test this in a live environment?

infra/dcp/modules/stack/variables.tf
```diff
@@ -69,15 +71,17 @@ variable "auth_config" {
...
+    enable                      = bool
+    instance_name               = string
+    memory_size_gb              = number
+    tier                        = string
+    location_id                 = string
+    alternative_location_id     = string
+    replica_count               = number
+    vpc_network_name            = string
+    vpc_connector_cidr          = string
+    vpc_connector_min_instances = optional(number, 2)
```
gmechali: can we change this to just be a number, you already set the default to 2/10 in the variables.

(I know we've left thispattern in other places but it's a mistake on our end :)

infra/dcp/terraform.tfvars.template
```diff
@@ -42,6 +42,10 @@ storage_create_artifacts_bucket = true
 # significantly improving performance. You may want to set this to true in production.
 enable_redis = false
 
+# VPC connector scaling for Redis. Lower min_instances reduces baseline cost.
```
gmechali: Can you revert this file enitrely?

We are intentionally trying to be very selective about which variables are in the template tfvars file. 

While all the variables you added should be usable by customers, I dont think they're high level enough to warrant being added to the template. 

We will absolutely need to update our docs to call out that these are some important optimization settings!

infra/dcp/variables.tf
```diff
@@ -151,6 +151,28 @@ variable "redis_vpc_connector_cidr" {
   default     = "10.13.0.0/28"
 }
 
+variable "redis_vpc_connector_min_instances" {
+  description = "Minimum number of VPC Access Connector instances. Lower values reduce baseline cost."
+  type        = number
+  default     = 2
+
+  validation {
+    condition     = var.redis_vpc_connector_min_instances >= 2 && var.redis_vpc_connector_min_instances <= 9
```
gmechali: I think the only thing we want to validate here is that min_instances < max_instances, which I see you have at the bottom!

But if I set my max_instances higher, than I dont see why we should restrict the min_instances to be between 2 and 9. 
They should be allowed to set these numbers to whatever they need.

saurabh-google:
> Thank you @saurabh-google !
> 
> Were you able to test this in a live environment?

I have tested it for a live deployment and I can see the Cloudrun > Revision History > Billing change to Request-based from Instance-based.

infra/dcp/modules/stack/variables.tf
saurabh-google: Done.

infra/dcp/terraform.tfvars.template
saurabh-google: Done.

infra/dcp/variables.tf
saurabh-google: Done.

PR #211, opened 2026-08-13
feat(datacommons-db): Add SpannerClient for schema versioning primitives

juliawu:
# Summary
Introduces `SpannerClient` in `packages/datacommons-db` to provide foundational Cloud Spanner primitives for DDL/DML/Query execution.

This is the first PR in a series intended to implement Schema Versioning and Migrations. Implementing the migration runner, migration scripts, and CLI updates will come in subsequent PRs.

#### 1. `SpannerClient` (`packages/datacommons-db/datacommons_db/client.py`)
- **Strict Identifier Validation:** Validates `project_id`, `instance_id`, `database_id`, and table names against strict regex patterns (`^[a-zA-Z0-9_\-]+$`).
- **Table Existence Checks:** `table_exists(table_name)` and `schema_version_table_exists()` query `information_schema.tables` via read snapshots.
- **Execution Primitives:**
  - `execute_ddl(ddl_statements)`: Supports single strings, semicolon-separated multi-statement strings, and lists of DDL statements. Waits on Spanner LROs (`operation.result()`).
  - `execute_dml(query, params, param_types)`: Executes read-write transactions with parameter binding.
  - `execute_query(query, params, param_types)`: Executes point-in-time snapshot reads.

#### 2. Package Interface (`packages/datacommons-db/datacommons_db/__init__.py`)
- Exported `SpannerClient` and `__version__` on top-level package namespace.

#### 3. Stateful In-Memory Test Suite (`packages/datacommons-db/tests/test_client.py`)
- Implemented `FakeSpannerDatabase`, `FakeSnapshot`, and `FakeTransaction` test doubles.
- 23 atomic unit tests covering identifier validation, table DDL, schema version ordering, DML transactions, and parameterized snapshot reads without mocking internal client methods.

#### 4. Documentation (`packages/datacommons-db/README.md`)
- Added usage documentation and code snippets for `SpannerClient`.

## Testing
- **Local Unit Tests:** `uv run pytest` -> All tests passed.
- **Linting:** `uv run ruff check` -> All checks passed.
- **Live Spanner Smoke Test:** Verified end-to-end on live GCP Spanner instance (`datcom-website-dev` / `dcp-testing` / `dev-juliawu-dc-db`) that each of the new client methods function as intended.


juliawu:
> ## Code Review
> This pull request introduces the `SpannerClient` utility class to manage Cloud Spanner schema migrations, DDL execution, and version tracking, along with comprehensive unit tests using an in-memory database test double. The reviewer feedback highlights several key improvement opportunities: making `project_id` optional to leverage GCP's automatic project detection, enhancing dialect compatibility for PostgreSQL-dialect Spanner databases, avoiding naive semicolon splitting of DDL statements, and providing a streaming alternative for queries to prevent loading large result sets entirely into memory.
> 

Good bot. I've agreed with and implemented each of your suggestions.

packages/datacommons-db/datacommons_db/clients/spanner_client.py
```diff
@@ -0,0 +1,249 @@
...
+
+def _validate_resource_id(name: str, value: object) -> None:
+    """Validate that a GCP / Spanner resource ID matches expected identifier patterns."""
+    if not isinstance(value, str) or not value or not _RESOURCE_ID_PATTERN.match(value):
+        raise ValueError(
+            f"Invalid {name} '{value}'. Must be a non-empty string containing only alphanumeric characters, underscores, and hyphens."
+        )
+
+
+class SpannerClient:
```
gmechali, line 80: this is the file for the spanner_client, can we name the file spanner_client? 

Also we have a subdirectory structure from here, with the /models subdir, I think it might be helpful to also move this to clients/spanner_client.py

Keeping a cleanly nested structure would be valuable for long term maintainability :)

packages/datacommons-db/datacommons_db/clients/spanner_client.py
```diff
@@ -0,0 +1,249 @@
...
+from google.auth.credentials import Credentials
+from google.cloud import spanner
+from google.cloud.spanner_v1.transaction import Transaction
+
+_SCHEMA_VERSION_TABLE_NAME = "SchemaVersion"
+_TABLE_NAME_PATTERN = re.compile(r"^[A-Za-z0-9_]+$")
+_RESOURCE_ID_PATTERN = re.compile(r"^[a-zA-Z0-9_\-]+$")
+
+
+def _validate_resource_id(name: str, value: object) -> None:
```
gmechali: imo this is a very usefl function to have, and it doesn't belong to the spanner client. 

This is a general helper. Could you make a new subdir for utils/ where we can hold similar helper functions?
@dwnoble should this be in datacommons-db? Should we have a datacommons-util package that gets looped in the CLI? With db/api/admin as diff packages, im not sure what's the right approach here. wdyt

packages/datacommons-db/datacommons_db/clients/spanner_client.py
```diff
@@ -0,0 +1,249 @@
...
+        database_id: str,
+        project_id: str | None = None,
+        credentials: Credentials | None = None,
+    ) -> None:
+        """Initialize the SpannerClient.
+
+        Args:
+            instance_id: Cloud Spanner instance ID.
+            database_id: Cloud Spanner database ID.
+            project_id: Optional GCP project ID. If None, auto-detected from the environment.
```
gmechali: ultra nit: might be nice to keep the ordering by scope --> project_id, instance_id, database_id

packages/datacommons-db/datacommons_db/clients/spanner_client.py
```diff
@@ -0,0 +1,249 @@
...
+
+        self.instance_id = instance_id
+        self.database_id = database_id
+
+        self.client = spanner.Client(project=project_id, credentials=credentials)
+        self.project_id = self.client.project
+        self.instance = self.client.instance(self.instance_id)
+        self.database = self.instance.database(self.database_id)
+
+    def table_exists(self, table_name: str) -> bool:
```
gmechali, line 110: nit: can we make this a list of tables? Would help avoid making multiple calls

packages/datacommons-db/datacommons_db/client.py
```diff
@@ -0,0 +1,249 @@
...
+        Args:
+            version: The integer schema version applied.
+            description: Description of the schema migration.
+        """
+        if not isinstance(version, int) or version < 0:
+            raise ValueError(f"Version must be a non-negative integer, got {version}")
+        if not isinstance(description, str) or not description.strip():
+            raise ValueError("Description must be a non-empty string.")
+
+        query = (
```
gmechali: Why hardcode this in the spanner client?

shouldn't this be our first application of a schema_migration script? where the migration script initalizes the SchemaVersion table?

packages/datacommons-db/datacommons_db/client.py
```diff
@@ -0,0 +1,249 @@
...
+        param_types = {"table_name": spanner.param_types.STRING}
+
+        with self.database.snapshot() as snapshot:
+            for _ in snapshot.execute_sql(
+                query, params=params, param_types=param_types
+            ):
+                return True
+            return False
+
+    def schema_version_table_exists(self) -> bool:
```
gmechali: This helper function is for one line of logic and only used once essentially adding clutter rather than removing.

On line 106, can we just do

 if not self.table_exists(_SCHEMA_VERSION_TABLE_NAME)
    return 0

packages/datacommons-db/datacommons_db/client.py
```diff
@@ -0,0 +1,249 @@
...
+        """
+        if not self.schema_version_table_exists():
+            return 0
+
+        # The version with the latest applied timestamp is the current version.
+        query = (
+            f"SELECT Version FROM {_SCHEMA_VERSION_TABLE_NAME} "  # noqa: S608
+            "ORDER BY AppliedTimestamp DESC LIMIT 1"
+        )
+        with self.database.snapshot() as snapshot:
```
gmechali: You basically seem to be rewriting execute_query_stream but in a non stream way here is that right? 

could you make a helper function for execute_query and have it self contain all of that? So moving forward all select calls can just be a one line call of this function?

packages/datacommons-db/datacommons_db/client.py
```diff
@@ -0,0 +1,249 @@
...
+            f"INSERT INTO {_SCHEMA_VERSION_TABLE_NAME} (Version, AppliedTimestamp, Description) "  # noqa: S608
+            "VALUES (@version, PENDING_COMMIT_TIMESTAMP(), @description)"
+        )
+        params = {"version": version, "description": description.strip()}
+        param_types = {
+            "version": spanner.param_types.INT64,
+            "description": spanner.param_types.STRING,
+        }
+
+        self.execute_dml(query, params=params, param_types=param_types)
```
gmechali: love the raising value errors above, I think we need to bring the same kind of error management to the returns here.
After every call to execute_dml/ddl/query we should be returning a success status and verifying that success state!

packages/datacommons-db/datacommons_db/clients/spanner_client.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def execute_ddl(self, ddl_statements: str | list[str]) -> None:
+        """Execute DDL statements and wait for completion.
+
+        Handles operations like CREATE TABLE, ALTER TABLE, etc.
+
+        Args:
+            ddl_statements: A single DDL statement string or a list of DDL statement strings.
+        """
+        statements: list[str] = []
```
gmechali: This looks a bit brittle to me. 

Ideally we shouldn't need to do any statement manipulation at this stage, that should all be done before we submit the schema change.

SQLGlot seems to allow us to validate DDL statements. Im not sure it's the best one, but it's an option.
I'd rather we put in some strong safeguards to ensure validity of the statement on the schema_migrations directory, than try to catch errors here :)

packages/datacommons-db/datacommons_db/clients/spanner_client.py
```diff
@@ -0,0 +1,249 @@
...
+
+        Handles operations like INSERT, UPDATE, DELETE
+
+        Args:
+            query: The parameterized DML statement.
+            params: Dictionary of parameters.
+            param_types: Dictionary of parameter types.
+
+        Returns:
+            The number of rows modified.
```
gmechali: I think we should return the status first, then maybe the number of rows, which you can do in python

packages/datacommons-db/datacommons_db/clients/spanner_client.py
juliawu: The Gemini-review bot pointed out that we should allow optional project_id to match the behavior of other client libaries. If project_id is optional, it has to go after the required arguments.

packages/datacommons-db/datacommons_db/clients/spanner_client.py
juliawu: Do you mean something like `tables_exist(self, table_names: list[str]): -> dict[str, bool]`? I find that interface clunky. I think the usual use case is a single-table existence check, so it becomes awkward to wrap it in a list. We could add `tables_exist(table_names: list[str]): -> dict[str, bool]` as it's own function to handle the multi-table case, or a `list_tables(): -> list[str]` function instead. Wdyt?

packages/datacommons-db/datacommons_db/clients/spanner_client.py
juliawu: Good idea. Moved to `clients/spanner_client.py`

packages/datacommons-db/datacommons_db/client.py
juliawu: Good call out. I've decided to remove the schema-specific functions from the client. As you've called out below as well, they belong in the migration-runner.

packages/datacommons-db/datacommons_db/client.py
juliawu: +1, will do that when this is implemented in the migration runner.

packages/datacommons-db/datacommons_db/client.py
juliawu: Good call out. Removed this in favor of adding it in the migration runner in a subsequent PR.

packages/datacommons-db/datacommons_db/clients/spanner_client.py
gmechali: I think in general it could make sense, but we actually have access to the Terraform with the actual project ID. I dont think we should ever rely on the env var bc of that.

If we start to support the CLI in a non-Terraform backed environment, then perhaps that can change. regardless this is a nit, so as you wish :)

juliawu: The point about having access to the terraform with project id has sold me. Ordered by scope!

packages/datacommons-db/datacommons_db/clients/spanner_client.py
juliawu: I've mad a new subdir for `utils/` for now, but I think having a datacommons-util package would be good eventually. Leaving the new package for a separate PR pending Dan's comments.

packages/datacommons-db/datacommons_db/clients/spanner_client.py
juliawu: Good point, I've removed the statement manipulation.

packages/datacommons-db/datacommons_db/clients/spanner_client.py
juliawu: Good idea about also returning status. Added explicit return classes for storing status and any error messages.

packages/datacommons-db/datacommons_db/client.py
juliawu: Good idea about the success status. As mentioned below, I've added a return class to explicitly track status and any error messages, and I'll use that in the migration runner layer.

packages/datacommons-db/datacommons_db/clients/spanner_client.py
```diff
@@ -0,0 +1,235 @@
...
+class DdlResult:
+    """Result of a DDL statement execution.
+
+    Attributes:
+        success: True if DDL statements completed successfully, False otherwise.
+        error: Error message string if execution failed, None otherwise.
+    """
+
+    success: bool
+    error: str | None = None
```
gmechali: I think this class should just have a status Enum + errorMessage str.

packages/datacommons-db/datacommons_db/clients/spanner_client.py
```diff
@@ -0,0 +1,235 @@
...
+class DmlResult:
+    """Result of a DML statement execution inside a read-write transaction.
+
+    Attributes:
+        success: True if the transaction committed successfully, False otherwise.
+        rows_affected: Number of rows modified by the DML statement (0 on failure).
+        error: Error message string if execution failed, None otherwise.
+    """
+
+    success: bool
```
gmechali: Same thing but Status + rows_affected + errorMessage

packages/datacommons-db/datacommons_db/clients/spanner_client.py
```diff
@@ -0,0 +1,235 @@
...
+            "WHERE table_schema IN ('', 'public') AND table_name = @table_name LIMIT 1"
+        )
+
+        params = {"table_name": table_name}
+        param_types = {"table_name": spanner.param_types.STRING}
+
+        result = self.execute_query(query, params=params, param_types=param_types)
+        return bool(result.success and result.rows)
+
+    def execute_ddl(self, ddl_statements: str | list[str]) -> DdlResult:
```
gmechali: why support both? Can we just take a list, and if you have a single one, you just need to wrap it in [MY_STATEMENT]

packages/datacommons-db/datacommons_db/clients/spanner_client.py
```diff
@@ -0,0 +1,235 @@
...
+            statements = [ddl_statements]
+        elif isinstance(ddl_statements, list):
+            statements = ddl_statements
+        else:
+            return DdlResult(
+                success=False,
+                error="ddl_statements must be a str or a list of str.",
+            )
+
+        if not statements:
```
gmechali: Could you move this to the first line of the function
have if not ddl_statements --> raise error

gmechali (commented):
Couple minor comments on the Result objects, but other than that LGTM. Thanks Julia!

packages/datacommons-db/datacommons_db/clients/spanner_client.py
juliawu: Switched to enum.

packages/datacommons-db/datacommons_db/clients/spanner_client.py
juliawu: Sg, switched to enum

packages/datacommons-db/datacommons_db/clients/spanner_client.py
juliawu: Originally wanted to mirror the fact that execute_query and execute_dml run single queries, but I don't feel strongly either way. Updated to only support a list.

packages/datacommons-db/datacommons_db/clients/spanner_client.py
juliawu: So, the original order here was because I thought if `ddl_statements = []`, then it should be treated as a successful no-op. Do we want to require non-empty strings here? That would mean callers would need some defensive boiler-plate if generating statements dynamically like:
```
if statements:
  result = client.execute_ddl(statements)
```

juliawu: After discussion, made empty statements an error.

gmechali (approved):
thank youu!!

packages/datacommons-db/datacommons_db/clients/spanner_client.py
gmechali: thank you! lgtm

